PR #3, opened 2017-06-27
STITCH-156 Add tests for Android SDK

jsflax:
Added Kotlin for testing.

Added RESTMock framework to mock network calls: https://github.com/andrzejchm/RESTMock .

Testing public methods of `StitchClient` class, as well as general authentication and pipeline flows.


.idea/markdown-exported-files.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
edaniels: Need this file?

build.gradle
```diff
@@ -15,6 +17,7 @@ buildscript {
 
 allprojects {
     repositories {
+        maven { url "https://jitpack.io" }
```
edaniels: What's this for?

.idea/misc.xml
```diff
@@ -37,7 +37,7 @@
     <ConfirmationsSetting value="0" id="Add" />
     <ConfirmationsSetting value="0" id="Remove" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
edaniels: Should find a way to stop this from changing :(

sdk/build.gradle
```diff
@@ -15,7 +16,8 @@ android {
         versionCode 20170505
         versionName "0.1.0"
 
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        multiDexEnabled true
```
edaniels: Will using `multiDexEnabled` affect apps importing the SDK or no?

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
+package com.mongodb.stitch
+
+import android.content.Context
```
edaniels: Looks like these imports need to be auto-organized

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+import io.appflate.restmock.android.AndroidAssetsFileParser
+import io.appflate.restmock.RESTMockServerStarter
+import io.appflate.restmock.utils.RequestMatchers.pathEndsWith
+
+/**
+ * Test public methods of StitchClient, running through authentication
+ * and pipeline flows.
+ */
+@RunWith(AndroidJUnit4::class)
+class StitchClientTest {
```
edaniels: Does this need a package access specifier or is it unnecessary?

edaniels: Oh this is Kotlin so the answer may very well be no

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+    companion object {
+        const val FakeAccessToken = "fake-access-token"
```
edaniels: is it a kotlin convention to uppercase the first letter of a constant?

edaniels:
cc: @mburdette if you want to skim it

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+                        "metadataFields" to arrayOf(
+                                FakeMetadataFieldEmail,
+                                FakeMetadataFieldBirthday
+                        )
+                )
+        )).toString()
+
+        /** Mock data for auth object from [StitchClient.logInWithProvider] */
+        val mockAuthData = JSONObject(mapOf(
+                "accessToken" to FakeAccessToken,
```
edaniels: Should we use constant values for these field names in general?

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+        )).toString()
+    }
+
+    /** Base context from test runner */
+    private val instrumentationCtx: Context = InstrumentationRegistry.getContext()
+    /** [StitchClient] for this test */
+    private var stitchClient: StitchClient? = null
+
+    @Before
+    fun setup() {
```
edaniels: so fun

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+        // start the mock server
+        RESTMockServerStarter.startSync(AndroidAssetsFileParser(instrumentationCtx), AndroidLogger())
+
+        // instantiate a new StitchClient using a dummy name and the mock baseUrl
+        val stitchClient = StitchClient(instrumentationCtx, "dummy-app", RESTMockServer.getUrl())
+        this.stitchClient = stitchClient
+
+        // clear all instances of the internal [SharedPreferences] to start with a clean slate
```
edaniels: What's this convention of using brackets around the class name?

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+        stitchClient.properties.clear()
+        stitchClient.javaClass.kotlin.declaredMemberProperties.first {
```
edaniels: Can you explain what's going on here?

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            it.name == "_preferences"
+        }.also { it.isAccessible = true }.get(stitchClient).also {
+            (it as SharedPreferences).edit().clear().commit()
+        }
+
+        // mock out all calls related to auth
+        RESTMockServer.whenPOST(pathContains("auth")).thenReturnString(mockAuthData)
```
edaniels: I'd like these to be more specific as spec-like such that we know exactly what type of path we are looking for. It may be easy to mock the wrong thing this way.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+        RESTMockServer.whenGET(pathEndsWith("auth")).thenReturnString(mockProviderData)
```
edaniels: Will these types of responses return JSON content-types or what?

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            }
+        })
+
+        // assert that we have not authenticated yet
+        assertThat(!stitchClient!!.isAuthenticated)
+
+        // fetch mocked authProviders
+        await(stitchClient!!.authProviders).let {
+            // assert all providers provided in mock are available
+            assertThat(it.hasAnonymous() && it.hasEmailPassword()
```
edaniels: If we have a test like this we should have another test where there is no provider available and a partial amount available to cover our edge cases

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            when (it) {
+                is GoogleAuthProviderInfo ->
+                    assertThat(it.clientId == FakeGoogleClientId)
+                is FacebookAuthProviderInfo ->
+                    assertThat(it.applicationId == FakeFacebookClientId)
+            }
+        }
+
+        // log in anonymously
+        await(stitchClient!!.logInWithProvider(AnonymousAuthProvider()))
```
edaniels: !! is similar to getting the optional and dereferencing?

sdk/src/androidTest/java/com/mongodb/stitch/TestUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Convenience method that wraps a [Task] in the [latch] method, awaiting the result.
+ *
+ * @param[task] pending task to wait for
+ *
+ * @return result of the task after it's been completed
+ */
+fun <T> await(task: Task<T>): T {
```
edaniels: so is using coroutines not a better idea with async and suspend?

.idea/misc.xml
jsflax: Will look into this. Could just ignore the file.

build.gradle
jsflax: This is for the RESTMock framework.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Unnecessary, public by default.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Looks like I was still thinking Swift-y here. Should be defaulting to Java naming conventions here: http://kotlinlang.org/docs/reference/coding-conventions.html will change.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: We could, but they would be unused at the moment, as we aren't going to be accessing the mock data by keys, but in their mapped object form.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Brackets is the equivalent of `{@link SharedPreferences}` in Java

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: I'm reflectively clearing out the `_properties` member of the StitchClient.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Unsure, but they are currently being handled correctly. If I want to return more precise responses, I have to instantiate a `MockResponse` and provide that information. Will change for precision.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Will change.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Correct, `!!` is the unsafe way to unwrap optionals. In this case, it's safe as we know 100% that the `stitchClient` will be non-null and unchanged.

sdk/src/androidTest/java/com/mongodb/stitch/TestUtils.kt
jsflax: https://kotlinlang.org/docs/reference/coroutines.html So after fiddling around with co-routines, I ran into a few issues, namely that they don't play well with the Tasks/Futures that we are currently using:

```kotlin
    inline fun <R> latch(count: Int = 1, block: CountDownLatch.() -> R): R {
        val latch = CountDownLatch(count)
        val r = latch.block()
        latch.await()
        return r
    }

    private suspend fun <T> fetch(task: Task<T>): T {
        latch { task.addOnCompleteListener { this.countDown() } }

        return task.result
    }

    @Test fun testAuth() = runBlocking {
        val deferredAuth = async(CommonPool) {
            fetch(stitchClient!!.logInWithProvider(AnonymousAuthProvider()))
        }

        val auth = deferredAuth.await()

        assert(auth.accessToken == FakeAccessToken)
    }
```

That was my experimentation. I ended up being able to accomplish all that in two lines without using coroutines:

```kotlin
@Test fun testAuth() {
        val auth = fetch(stitchClient!!.logInWithProvider(AnonymousAuthProvider()))

        assertThat(auth.accessToken == FakeAccessToken)
}
```

It looks like to take advantage of coroutines, we'd need to write our own Promise class in kotlin that uses the `suspended` keyword. It's possible we can implement our own `Task` class (`Task` is abstract after all), but that seems like overkill right now.

Could be a cool thought for a side project. Kotlin promises would be nice while the language is still in its nascent stages.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Will fix.

sdk/build.gradle
jsflax: I'm going to have to test this against a sample app before I can properly answer this question.

sdk/src/androidTest/java/com/mongodb/stitch/StitchClientTest.kt
jsflax: Should also probably make sure the `StitchClient`is catching error responses (for mock errors) while we're at it. Will add.

sdk/src/androidTest/java/com/mongodb/stitch/TestUtils.kt
edaniels: makes sense to not do it then

jsflax:
@edaniels PTAL. I've tested this against MongoCipher after a local deploy to ensure that multidex/kotlin will not have negative effects. The only change I didn't tend to was changing that `path` mocks. Thanks to the HTTP verb-age, there will not be any conflicts... for now. But it's something we'll need to keep an eye on.

PR #5, opened 2017-07-12
Swapped out bson snapshot for alpha1

jsflax:


sdk/build.gradle
```diff
@@ -75,8 +75,8 @@ dependencies {
     compile 'com.google.android.gms:play-services-base:10.2.1'
     compile "com.google.android.gms:play-services-gcm:10.2.1"
 
-    // TODO: Remove SNAPSHOT once BSON 3.5.0 is released
-    compile 'org.mongodb:bson:3.5.0-SNAPSHOT'
+    // TODO: Remove -alpha1 once BSON 3.5.0 is released
```
edaniels: This TODO is no longer relevant since we will have to do more work to go straight to BSON 3.5.0 due to Extended JSON Spec incompatibilities from  3.5.0-alpha1 to 3.5.0

PR #6, opened 2017-07-12
Allow SDK to be built via command-line as long as ANDROID_HOME is set

edaniels:
Repurposed `gradle-mvn-push.gradle`


jsflax:
LGTM 👍

PR #7, opened 2017-10-10
STITCH-719 Upgrade Android SDK to latest BSON version

jsflax:


edaniels:
Note to verify details publishing to different repo and that tests are still working

PR #9, opened 2017-10-23
STITCH-548: Proactively refreshing access token

haleyowen:


sdk/src/main/java/com/mongodb/stitch/android/StitchClient.java
```diff
@@ -775,6 +782,11 @@ public void onErrorResponse(final VolleyError error) {
         return future.getTask();
     }
 
+    private boolean isTokenExpired(final String jwt) {
```
jsflax: [minor] This method may be better suited be a public method within the `DecodedJWT` class. And since these tokens actually are JWTs, we might as well change the type of the tokens (within the `Auth` class)  to be typed as `DecodedJWT`s to better represent their state.

sdk/src/main/java/com/mongodb/stitch/android/auth/DecodedJWT.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+
+@SuppressWarnings("WeakerAccess")
+public class DecodedJWT {
+
+    private final String[] parts;
+    private final Map<String, Object> payload;
+
+    public DecodedJWT(final String jwt) {
+        parts = splitToken(jwt);
```
jsflax: [rec] Seems like once we've mapped the payload to our field, we no longer need the `parts` field to be stored. It would be more useful for a user to have the raw jwt saved as a field as such:

```java
public final String raw;
...
public DecodedJWT(final String jwt) {
    this.raw = jwt;
    ...
}
```

PR #19, opened 2018-01-08
STITCH-1095 Google and Facebook Auth providers broken in Android SDK

jsflax:


jsflax:
Note– tests for OAuth are non-trivial to write. They will be added in STITCH-998.

PR #25, opened 2018-02-06
STITCH-1178 Set default volley timeout to 15 seconds

adamchel:
This change sets the default volley timeout for requests from StitchClient to 15 seconds by always using a `DefaultRetryPolicy` with a 15000ms timeout when enqueuing requests.

Didn't add a test yet since I wasn't sure how we would do it without mocking long response times.

We could also add a method for users to set the default timeout on a `StitchClient`. let me know if we want to add this (May also be worthwhile to think about as something to standardize across the SDKs).

### DIFF
--- a/stitch/src/main/java/com/mongodb/stitch/android/StitchClient.java
+++ b/stitch/src/main/java/com/mongodb/stitch/android/StitchClient.java
@@ -745,7 +745,7 @@ public class StitchClient {
      *
      * @param request The request to be added to the queue.
      */
-    private void enqueueRequest(Request request) {
+    private void enqueueRequest(final Request request) {
         request.setTag(this);
         request.setRetryPolicy(new DefaultRetryPolicy(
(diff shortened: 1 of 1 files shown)


stitch/src/main/java/com/mongodb/stitch/android/StitchClient.java
```diff
@@ -747,6 +739,21 @@ String getAuthProvidersLoginRoute(String providerType) {
         private final String USERPASS_RESET_SEND = USERPASS + "reset/send";
     }
 
+    /**
+     * Adds a cancellable request to the Volley queue for this StitchClient. The request will
+     * timeout if it does not complete in DEFAULT_TIMEOUT_MS milliseconds.
+     *
+     * @param request The request to be added to the queue.
+     */
+    private void enqueueRequest(Request request) {
```
edaniels: [nit] `final` request

PR #34, opened 2018-04-27
STITCH-1317 Handle transport timeouts

adamchel:
This was basically the same change as for Swift. I added a transport timeout setting to client configuration, and I modified doRequest in the StitchRequestClient to perform the Transport round trip on a background thread, and wrote a test. 

I'm somewhat concerned with the performance implications of the ExecutorService I'm using to wait for the round trip to finish, so I would like some feedback on that, mainly since I'm not too familiar with concurrency in Java.

### DIFF
--- a/core/src/main/java/com/mongodb/stitch/core/internal/net/StitchRequestClient.java
+++ b/core/src/main/java/com/mongodb/stitch/core/internal/net/StitchRequestClient.java
@@ -6,15 +6,26 @@ import com.mongodb.stitch.core.StitchRequestException;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 public class StitchRequestClient {
 
   private final String baseURL;
   private final Transport transport;
+  private final Long transportTimeout;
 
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/mongodb/stitch/core/internal/net/StitchRequestClient.java
```diff
@@ -27,11 +38,39 @@ private static Response inspectResponse(final Response response) {
   }
 
   public Response doRequest(final StitchRequest stitchReq) {
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
adamchel: I'm concerned about this line in particular. Is there going to be significant overhead in creating a thread that will only live for a single network request? Would it be better if this request executor thread was a member of the request client so it's shared among requests? Or even perhaps to have a thread pool?

adamchel: I think having this `ExecutorService` be an instance or static member of `StitchRequestClient` would be most similar to what I have in the Swift SDK (as the PR currently stands) but I'm open to suggestions.

jsflax: Many threads are being spun up for various tasks during the lifecycle of any app. However, I'm not sure this is how we should be doing it, but not because of overhead. 

Either we just spin up a single thread and don't bother with executors, or spin up a thread pool.

adamchel:
Trying to rebase on eric's recent work on v4-alpha is a mess of merge conflicts. Since I'm going to be pretty much redoing this anyway, I'm going to close this and open a new PR when it's ready.

PR #42, opened 2018-05-11
STITCH-1317 Handle Transport timeouts

adamchel:
This is pretty much the same as the iOS changes, though 

* the tests were slightly different
* there were more `callFunction` variants because of the codecs
* I added `@Nullable` in a few places to make it clear to us that a particular type would often be validly `null`. Let me know if there's another way I should be doing this. In iOS this was just an optional but we can't do that here.


core/src/test/java/com/mongodb/stitch/core/testutil/Constants.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.stitch.core.testutil;
+
+public final class Constants {
+  public static final Long DEFAULT_TRANSPORT_TIMEOUT_MILLISECONDS = 15000L;
+}
```
jsflax: [nit] missing newline. Also I'm not a fan of a Constants file, but it seems fine here.

PR #43, opened 2018-05-14
STITCH-1369 - Implement AWS SES Service in Core, Android, and Server

edaniels:


edaniels:
One sec forgot the SendResult

android/services/aws-ses/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest package="com.mongodb.stitch.android.services.twilio">
```
jsflax: [minor] this seems wrong

core/services/aws-ses/src/main/java/com/mongodb/stitch/core/services/aws/ses/CoreAwsSesServiceClient.java
```diff
@@ -0,0 +1,44 @@
...
+    this.service = service;
+  }
+
+  protected void sendEmailInternal(
+      final String toAddress,
+      final String fromAddress,
+      final String subject,
+      final String body
+  ) {
+    final Document args = new Document();
```
jsflax: [minor] if I'm being a stickler, we can pull these keys out into consts, or, create a codec for this structure.

android/services/aws-ses/src/main/AndroidManifest.xml
edaniels: it is!

core/services/aws-ses/src/main/java/com/mongodb/stitch/core/services/aws/ses/CoreAwsSesServiceClient.java
edaniels: I think they don't really belong anywhere but here if they're used just once.

jsflax: Alright, disregard then

PR #46, opened 2018-05-18
STITCH-1448 Implement User API Key Client

adamchel:
This is mostly similar to the iOS PR, though Java type erasure made exposing the provider clients in StitchAuth not quite the same as iOS. We couldn’t use the name `getProviderClient` for both authenticated and unauthenticated provider clients, so there’s a `getAuthenticatedProviderClient`.

It was also a little funky to decode a list of User API keys for fetchApiKeys. I had to use `com.fasterxml.jackson.core.type.TypeReference;` and create a separate decode function for lists. If you guys have any better ideas let me know.


android/core/src/main/java/com/mongodb/stitch/android/core/auth/StitchAuth.java
```diff
@@ -20,20 +20,33 @@
...
 public interface StitchAuth extends Closeable {
+  /**
+   * Gets an authenticated client for the given authentication provider. Most authentication
+   * providers will allow creation of a client without a name of the provider.
+   *
+   * @param provider The provider that will create a client for the authentication provider.
+   * @param <ClientT> The type of client to be returned by the provider.
+   * @return A client to interact with the authentication provider.
+   */
+  <ClientT> ClientT getAuthenticatedProviderClient(
```
edaniels: I'd really like to avoid having an extra method like this with this kind of name. There should be another way to do this. If the type erasure makes it impossible or just the lack of the `where` constraints, can we just make the factories accept both a requestclient and authrequestclient. If it goes unused it's fine. The users of this SDK won't care.

core/sdk/src/main/java/com/mongodb/stitch/core/auth/providers/userapikey/CoreUserApiKeyAuthProviderClient.java
```diff
@@ -16,10 +16,142 @@
...
+   * Fetches the user API keys associated with the current user.
+   */
+  protected List<UserApiKey> fetchApiKeysInternal() {
+    final StitchAuthRequest.Builder reqBuilder = new StitchAuthRequest.Builder();
+    reqBuilder
+            .withMethod(Method.GET)
+            .withPath(this.getBaseRoute())
+            .withRefreshToken()
+            .withShouldRefreshOnFailure(false);
+    return decodeList(
```
edaniels: If you add the ability to do decoding with `Decoder<T>` to doAuthenticatedRequest like the Json variants, you can just use this which is in my current branch:
```
class CollectionDecoder<TResult> implements Decoder<Collection<TResult>> {

  private final Decoder<TResult> decoder;

  public CollectionDecoder(final Decoder<TResult> decoder) {
    this.decoder = decoder;
  }

  @Override
  public Collection<TResult> decode(final BsonReader reader, final DecoderContext decoderContext) {
    final Collection<TResult> docs = new ArrayList<>();
    reader.readStartArray();
    while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
      final TResult doc = decoder.decode(reader, decoderContext);
      docs.add(doc);
    }
    reader.readEndArray();
    return docs;
  }
}
```

e.g.
```
return service.callFunctionInternal(
        "find",
        Collections.singletonList(args),
        new CollectionDecoder<>(decoder));
```

core/sdk/src/main/java/com/mongodb/stitch/core/auth/providers/CoreAuthProviderClient.java
```diff
@@ -16,39 +16,82 @@
...
+          final String providerName,
+          final RequestClientT requestClient,
+          final String baseRoute) {
     this.providerName = providerName;
     this.requestClient = requestClient;
-    this.authRoutes = authRoutes;
+    this.baseRoute = baseRoute;
+  }
+
+  /**
```
edaniels: [remove] Hopefully you can remove all of this with my decoder suggestion.

core/sdk/src/main/java/com/mongodb/stitch/core/auth/providers/userapikey/CoreUserApiKeyAuthProviderClient.java
```diff
@@ -16,10 +16,142 @@
...
+  /**
+   * Fetches the user API keys associated with the current user.
+   */
+  protected List<UserApiKey> fetchApiKeysInternal() {
+    final StitchAuthRequest.Builder reqBuilder = new StitchAuthRequest.Builder();
+    reqBuilder
+            .withMethod(Method.GET)
+            .withPath(this.getBaseRoute())
+            .withRefreshToken()
+            .withShouldRefreshOnFailure(false);
```
edaniels: [nit] Can exclude each `withShouldRefreshOnFailure` since the builder is smart enough to know using a refresh token means you cannot refresh.

server/core/src/main/java/com/mongodb/stitch/server/core/auth/internal/StitchAuthImpl.java
```diff
@@ -50,7 +53,17 @@ public StitchAuthImpl(
   }
 
   @Override
-  public <T> T getProviderClient(final AuthProviderClientFactory<T> provider) {
+  public <ClientT> ClientT getAuthenticatedProviderClient(
+          final AuthProviderClientFactory<ClientT, StitchAuthRequestClient> provider) {
+    if (!isLoggedIn()) {
```
edaniels: [minor] We should always let these be created. You may log out while you have one anyway. You don't need an authenticated auth request client to get here anyway.

adamchel: The same change will need to be made on iOS, I'll file a ticket since I don't want to touch iOS until Jason is done with it

android/core/src/main/java/com/mongodb/stitch/android/core/auth/StitchAuth.java
adamchel: I'm going with the factory accepting both types of request clients. I could not figure out another way to do it given the type erasure.

PR #61, opened 2018-09-11
STITCH-1961 Implement Change Event Coalescence Algorithm

jsflax:


edaniels:
@adamchel for visibility

core/services/mongodb-sync/src/main/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizer.java
```diff
@@ -1249,10 +1249,18 @@ public DeleteResult deleteOneById(
       return DeleteResult.acknowledged(0);
     }
 
-    final DeleteResult result = getLocalCollection(namespace)
+      final DeleteResult result = getLocalCollection(namespace)
         .deleteOne(getDocumentIdFilter(documentId));
     final ChangeEvent<BsonDocument> event =
         changeEventForLocalDelete(namespace, documentId, true);
+
+    if (config.getLastUncommittedChangeEvent() != null &&
```
edaniels: [nit] Add a short and sweet comment about why we do this.

core/services/mongodb-sync/src/main/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizer.java
```diff
@@ -1249,10 +1249,18 @@ public DeleteResult deleteOneById(
       return DeleteResult.acknowledged(0);
     }
 
-    final DeleteResult result = getLocalCollection(namespace)
+      final DeleteResult result = getLocalCollection(namespace)
```
edaniels: Remove the space?

core/services/mongodb-sync/build.gradle
```diff
@@ -15,6 +17,17 @@ dependencies {
     implementation project(':core:core-services:stitch-core-services-mongodb-remote')
 
     api('org.mongodb:mongodb-driver-embedded:3.8.1')
+
+    testImplementation project(':core:stitch-core-admin-client')
+    testImplementation project(':core:stitch-core-testutils')
+
+    testImplementation project(':core:core-services:stitch-core-services-mongodb-local')
```
edaniels: [q] Do you actually need these still?

core/services/mongodb-sync/src/test/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                ),
+                Mockito.spy(CoreRemoteMongoClientImpl(Mockito.mock(CoreStitchServiceClient::class.java))),
+                networkMonitor,
+                authMonitor
+        )
+    }
+
+    @After
+    fun teardown() {
+        val dataSpy = Mockito.spy(dataSynchronizer!!)
```
edaniels: Seems odd to keep spying on this in each test and have the !!s everywhere. You could take the same approach as the integration tests to make it read a little better and spy it and !! it just once too.

core/services/mongodb-sync/src/test/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Test
+    fun testInsertOneAndSync() {
+        val dataSpy = Mockito.spy(dataSynchronizer!!)
+
+        insertOneAndSync(dataSpy, BsonDocument())
+
+        assertEquals(dataSpy.getSynchronizedDocuments(namespace).first().documentId, insertedId)
+        assertEquals(dataSpy.getSynchronizedDocuments(namespace).size, 1)
```
edaniels: If you make an assertion like this, then you should assert prior to the insert that it's size is 0. I would also check that size is 1 before doing first, otherwise first may not have anything.

edaniels: I would also fill out the BsonDocument, and do a find one on it to verify what you inserted is there. In addition to this, this test doesn't feel complete. The data synchronizer will eventually try to push this to the remote. You could check this by running a sync pass and mock out the insertion for the remote. If you do this, you prove that the change event in the config is of type INSERT (which could also check but this would make it unnecessary)

core/services/mongodb-sync/src/test/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -0,0 +1,151 @@
...
+    fun testDeleteOneById() {
+        val dataSpy = Mockito.spy(dataSynchronizer!!)
+
+        insertOneAndSync(dataSpy, BsonDocument())
+
+        assertEquals(dataSpy.getSynchronizedDocuments(namespace).size, 1)
+
+        dataSpy.deleteOneById(namespace, insertedId)
+
+        assertEquals(dataSpy.getSynchronizedDocuments(namespace).size, 0)
```
edaniels: Similar comments about running the same lifecycle flow

core/services/mongodb-sync/src/test/java/com/mongodb/stitch/core/services/mongodb/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+        assertTrue(dataSpy.getSynchronizedDocuments(namespace).size > 0)
+
+        val namespaceArg = ArgumentCaptor.forClass(MongoNamespace::class.java)
+        val idArg = ArgumentCaptor.forClass(BsonValue::class.java)
+        val eventListenerArg = ArgumentCaptor.forClass(changeEventListener::class.java)
+        val documentCodecArg = ArgumentCaptor.forClass(BsonDocumentCodec::class.java)
+
+        verify(dataSpy).watchDocument(
```
edaniels: This seems like an odd thing to watch. This is reaching too far into internal details. I don't think this block of code is really necessary.

jsflax: I do this to capture the id.

PR #63, opened 2018-09-18
STITCH-1952 Hook up irrecoverable errors to error listener

jsflax:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ErrorListener.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface ErrorListener {
+
+  /**
+   * Called when an error happens for the given document id.
+   *
+   * @param documentId the _id of the document related to the error.
+   * @param error the error.
+   */
+  void onError(final BsonValue documentId, final Error error);
```
edaniels: [minor] Let's use Exception. An Error is really for classes of errors that require the program to terminate.

edaniels: Throwable is also an option but starting at Exception lets us move up to Throwable in the future.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -599,8 +616,14 @@ private void syncLocalToRemote() {
 
           case REPLACE: {
             if (localDoc == null) {
-              throw new IllegalStateException(
-                  "expected document to exist for local replace change event: %s");
+              IllegalStateException illegalStateException = new IllegalStateException(
```
edaniels: final

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -516,6 +521,18 @@ private void syncRemoteChangeEventToLocal(
         remoteChangeEvent);
   }
 
+  private void emitError(BsonValue docId, String msg) {
```
edaniels: [minor] Similar to emitting events, we should use the dispatcher for this so that we can continue on with our code. Right now we can block forever depending on what the user's callback does.

jsflax:
@edaniels PTAL

PR #64, opened 2018-09-20
STITCH-1962 Rename uncomitted writes field in change events

jsflax:


jsflax:
Also includes lint fixes

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/ChangeEvent.java
```diff
@@ -94,7 +94,7 @@ public UpdateDescription getUpdateDescription() {
     return updateDescription;
   }
 
-  public boolean isLocalWritePending() {
+  public boolean hasUncommittedWrites() {
     return localWritePending;
```
adamchel: [nit] should we change the field name as well?

jsflax: sure

PR #68, opened 2018-10-11
STITCH-1954 Implement instance id and counter versioning scheme

adamchel:
This is complete to the best of my knowledge and based on offline conversations I've had with Eric, but I threw in some comments where I'm still unsure if stuff needs to be done or if it should be delegated to future tickets.

The tests from 1950 (specifically `testConfigure`) are still flaky, but otherwise everything passes.

I'll keep an eye on evergreen as well.


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DocumentVersionInfo.java
```diff
@@ -19,57 +19,181 @@
...
+
+  /**
+   * Returns whether this version is non-empty (i.e. a version from a document with no version)
+   * @return true if the version is non-empty, false if the version is empty.
+   */
+  boolean isNonEmptyVersion() {
+    return versionValues != null;
+  }
+
+  private @Nonnull VersionValues getVersionValuesThrowIfNull() {
```
jsflax: [minor] Using `ThrowIfNull` as a suffix here, and the exception thrown here, is arguably a bit of an anti-pattern. If this _is_ an illegal state, we should throw that exception in the constructor– otherwise, it's on the user to call the `#isNonEmptyVersion` check.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DocumentVersionInfo.java
```diff
@@ -19,57 +19,181 @@
...
+  private static final String INSTANCE_ID_FIELD = "id";
+  private static final String VERSION_COUNTER_FIELD = "v";
+
+  private DocumentVersionInfo(
+          @Nullable final BsonDocument version,
+          @Nullable final BsonValue documentId
+  ) {
+    if (version != null) {
+      this.versionDoc = version;
+      this.versionValues = new VersionValues(
```
jsflax: [minor][related to comment below] We should be checking that these keys exist on the `versionDoc` before reading them.

If it the doc is null _or_ they do not exist _and_ this is a disallowed state, we should throw an `IllegalStateException`.

jsflax: From what I can tell here, if the `versionDoc` is null, that _should_ be an OK state, and is covered by `isNonEmptyVersion`. Maybe the exception should be if the versionDoc is malformed? Just food for thought– I think to an extent, there's a stylistic choice here.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1355,24 +1390,34 @@ private void replaceOrUpsertOneFromResolution(
...
             getDocumentIdFilter(documentId),
             docToReplace,
             new FindOneAndReplaceOptions().upsert(true).returnDocument(ReturnDocument.AFTER));
     final ChangeEvent<BsonDocument> event;
+
+    // TODO(QUESTION FOR REVIEW): it seems like we are setting pending writes at the
+    // old version, rather than the new version. Is this intentional? It seems like we should be
+    // setting the pending writes at the newly generated version. I'm confused though because if I
+    // use the new version here instead, tests fail because updates don't end up happening on a
+    // sync pass.
```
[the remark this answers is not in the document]
jsflax: I'm also thinking it could be nice to have this tested as a unit test.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -525,14 +523,43 @@ private void syncRemoteChangeEventToLocal(
...
+
+      // TODO(QUESTION FOR REVIEW): The testFrozenDocumentConfig test fails if I call this, but
+      // other tests pass. Shouldn't the receipt of this document indicate that this is the last
+      // known remote version of this document?
+      // docConfig.setLastKnownRemoteVersion(lastKnownRemoteVersion);
+
+      // TODO(QFR): I ask because when we are looking at hasCommittedVersion, we are using the
+      // lastKnownRemoteVersion of the doc config rather than the last known remote version based
+      // on the incoming change event we got here, which would make this not conform to spec.
+      if (docConfig.hasCommittedVersion(localVersion)) {
```
edaniels: [minor] This should just be remote version

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -525,14 +523,43 @@ private void syncRemoteChangeEventToLocal(
     final BsonDocument docFilter = getDocumentIdFilter(docConfig.getDocumentId());
     final BsonDocument localDocument = localColl.find(docFilter).first();
 
-    final BsonValue lastKnownRemoteVersion;
+    final BsonDocument lastKnownRemoteVersion;
```
edaniels: if it helps, name this `currentRemoteVersion` which will become `lastKnownRemoteVersion` in the config

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -525,14 +523,43 @@ private void syncRemoteChangeEventToLocal(
...
+                        logicalT,
+                        nsConfig.getNamespace(),
+                        docConfig.getDocumentId(),
+                        remoteVersionInfo.getSyncProtocolVersion()));
 
-      if (docConfig.hasCommittedVersion(lastKnownRemoteVersion)) {
+        return;
+      }
+
+      // TODO(QUESTION FOR REVIEW): The testFrozenDocumentConfig test fails if I call this, but
```
edaniels: There reason not to set this and why you can get rid of that set method is this would always make the next statement true.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
edaniels: discussed offline. this is correct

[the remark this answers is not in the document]
jsflax: 👍

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DocumentVersionInfo.java
adamchel: Definitely agree that this class is a little funky, I refactored it a little bit per your comments and eric's comments to make it harder to reach an illegal state.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -525,14 +523,44 @@ private void syncRemoteChangeEventToLocal(
...
+                        remoteChangeEvent.getOperationType().toString()));
 
-      if (docConfig.hasCommittedVersion(lastKnownRemoteVersion)) {
+        return;
+      }
+
+      // TODO(QUESTION FOR REVIEW): The testFrozenDocumentConfig test fails if I call this, but
+      // other tests pass. Shouldn't the receipt of this document indicate that this is the last
+      // known remote version of this document?
+      // docConfig.setLastKnownRemoteVersion(lastKnownRemoteVersion);
```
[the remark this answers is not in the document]
adamchel: no longer relevant, worked it out with Eric and we know have the correct behavior

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1345,7 +1380,7 @@ private void replaceOrUpsertOneFromResolution(
       final MongoNamespace namespace,
       final BsonValue documentId,
       final BsonDocument document,
-      final BsonValue atVersion,
+      final BsonDocument atVersion,
```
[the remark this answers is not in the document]
adamchel: This is because the consumer of `atVersion` is only operating on the pure BSON of the version, whereas `DocumentVersionInfo` is an abstraction on top of that that we use for various other checks.

`DocumentVersionInfo` is allowed to represent an empty version, meaning consumers of atVersion would then have to check for those kinds of states. I think this is fine for now, especially since consumers of `atVersion` never convert back to a `DocumentVersionInfo`, and they don't have any use for its methods/functionality

adamchel:
@edaniels @jsflax this is ready for another look. On top of resolving your feedback, I also did some general cleanup in `DataSynchronizer.java` to get rid of redundant code and code that was made unnecessary with some of the other refactors that we made

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DocumentVersionInfo.java
```diff
@@ -94,4 +230,21 @@ static BsonDocument getVersionedFilter(
...
     return filter;
   }
+
+  /**
+   * Given a DocumentVersionInfo, returns a BSON document representing the next version. This means
+   * and incremented version count for a non-empty version, or a fresh version document for an
+   * empty version.
+   * @return a BsonDocument representing a synchronization version
+   */
+  BsonDocument getNextVersion() {
```
edaniels: [nit] Add a unit test(s) for this

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
jsflax: 👍

jsflax:
Could use a unit test for the logic we spoke about earlier in this PR.

PR #69, opened 2018-10-12
STITCH-1956 Fill in missing unit tests for every relevant task and clean exisiting tests.

jsflax:
*Do not review.*

This is for agreement purposes only.


core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,60 +96,849 @@ class DataSynchronizerUnitTests {
...
+
+            dataSynchronizer.doSyncPass()
+
+            emitEventSemaphore.acquire(1)
+
+            verify(changeEventListener, times(4)).onEvent(eq(document1["_id"]), any())
+            verify(conflictHandler, times(1)).resolveConflict(eq(document1["_id"]), any(), any())
+            verify(errorListener, times(0)).onError(eq(document1["_id"]), any())
+
+            // verify we are inserting!
```
adamchel: [nit] verify we are not inserting? or that we desync-ed/froze?

adamchel: or is this verifying that the conflict resolver deleted the document?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,60 +96,849 @@ class DataSynchronizerUnitTests {
...
+
+            dataSynchronizer.doSyncPass()
+
+            emitErrorSemaphore.acquire()
+
+            verify(changeEventListener, times(1)).onEvent(eq(document2["_id"]), any())
+            verify(conflictHandler, times(0)).resolveConflict(eq(document2["_id"]), any(), any())
+            verify(errorListener, times(1)).onError(eq(document2["_id"]), any())
+
+            assertEquals(
```
adamchel: [minor] I think in general, for all these tests, we should add a comment when checking the document in the local collection to explain why we're checking for the specific expected state. I was confused at first why we were expecting the document to exist when the insert failed, but I realized it's because we expect the document to be there locally, just not remotely.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,60 +96,849 @@ class DataSynchronizerUnitTests {
...
+
+            dataSynchronizer.doSyncPass()
+
+            emitEventSemaphore.acquire(1)
+
+            verify(changeEventListener, times(4)).onEvent(eq(document1["_id"]), any())
+            verify(conflictHandler, times(1)).resolveConflict(eq(document1["_id"]), any(), any())
+            verify(errorListener, times(0)).onError(eq(document1["_id"]), any())
+
+            // verify we are inserting!
```
adamchel: inserting the resolution to the conflict?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -187,4 +993,128 @@ class DataSynchronizerUnitTests {
...
+
+        work(dataSynchronizer, collectionMock)
+    }
+
+    companion object {
+        private fun newDoc(key: String = "hello", value: BsonValue = BsonString("world")): BsonDocument {
+            return BsonDocument("_id", BsonObjectId()).append(key, value)
+        }
+
+        private fun withoutVersionId(document: BsonDocument?): BsonDocument? {
```
adamchel: [heads up] this should be renamed to `withoutSyncVersion` to be consistent with the change I made in my PR

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,60 +96,849 @@ class DataSynchronizerUnitTests {
...
+        networkMonitor.isOnline = false
+
+        dataSynchronizer.onNetworkStateChanged()
+
+        verify(dataSynchronizer, times(1)).start()
+        verify(dataSynchronizer, times(1)).stop()
     }
 
     @Test
     @Suppress("UNCHECKED_CAST")
```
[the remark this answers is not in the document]
jsflax: Removes unnecessary warnings due to us having to cast certain mocks.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -340,7 +340,7 @@ public void swapSyncDirection(final boolean localToRemoteFirst) {
    * @return whether or not the synchronization pass was successful.
    */
   public boolean doSyncPass() {
-    if (!syncLock.tryLock()) {
+    if (!this.isConfigured || !syncLock.tryLock()) {
```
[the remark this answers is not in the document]
jsflax: No, this is actually how the method should function. I don't believe this will solve any races.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,60 +96,849 @@ class DataSynchronizerUnitTests {
...
+            `when`(coreRemoteMongoCollectionMock.insertOne(any())).thenThrow(
+                StitchServiceException("E11000", StitchServiceErrorCode.MONGODB_ERROR)
+            )
 
-        // unfreeze the configs and reload
-        dataSynchronizer.getSynchronizedDocuments(namespace).forEach {
-            it.isFrozen = false
+            dataSynchronizer.insertOneAndSync(namespace, document2)
+
+            assertEquals(
```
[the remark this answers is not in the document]
jsflax: @edaniels Could you be a bit more specific with #2? Do you mean a completely unsupported operation type?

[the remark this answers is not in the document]
edaniels: Right.

jsflax:
This is a large PR.

- Add several tests for each of our core models (configs, ChangeEvents, etc.). These are mostly sanity checks and round trip checks.
- Add a suite of unit tests to test the ins and outs of the `DataSynchronizer`, `NamespaceChangeStreamListener`, and `CoreSync`. These tests are managed by the `SyncTestHarness`, which contains a series of helper methods in `DataSynchronizerTestContext`. The test context should make testing easier to read and more manageable.
- Consolidate android/server test logic. Proxy relevant methods through a core proxy and run the tests in core. See `ProxyRemoteMethods`, `ProxySyncMethods`, `SyncIntTestProxy`, `SyncIntTestRunner`.
- Add comments to integration tests, making them more legible.
- Fix `testConfigure` race in Int tests.
- Various bug fixes.

@edaniels @adamchel  PTAL.

Note: Seems like GitHub is having some issues. This is the patch for the latest commit on evg: https://evergreen.mongodb.com/version/5bcdf2dd2a60ed479842434d

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestRunner.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    @Test
+    fun testStaleFetchSingle()
+
+    @Test
+    fun testStaleFetchSingleDeleted()
+
+    @Test
+    fun testStaleFetchMultiple()
+}
```
edaniels: EOF

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -0,0 +1,1361 @@
...
+                hasChangeEventListenerBeenInvoked = true
+                changeEventListenerSemaphore.release()
+            },
+            ErrorListener { _, _ -> }
+        )
+
+        // insert a document remotely, and wait for the event to store
+        val sem = watchForEvents(syncTestRunner.namespace)
+        remoteColl.insertOne(Document("_id", insertedId).append("fly", "away"))
+        sem.tryAcquire(3, TimeUnit.SECONDS)
```
edaniels: [mahor] none of these tests that previously had just a acquire should have try acquire now. That encourages races to occur. If 3 seconds pass and nothing happens, this proceeds.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncTestHarness.kt
```diff
@@ -0,0 +1,567 @@
...
+            dataSynchronizer.configure(
+                namespace,
+                conflictHandler,
+                changeEventListener,
+                errorListener,
+                bsonDocumentCodec)
+        }
+
+        override fun waitForEvent() {
+            eventSemaphore?.tryAcquire(2, TimeUnit.SECONDS)
```
edaniels: These should all change too. We shouldn't base any of these on time

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncTestHarness.kt
```diff
@@ -0,0 +1,567 @@
+package com.mongodb.stitch.core.services.mongodb.remote.sync.internal
```
edaniels: Should this have the word Unit in it to denote that it's only for unit tests?

jsflax: I was under the impression that "harness" was a term generally for unit tests. Though Int tests have setup methods, you wouldn't want a harness per se.

On the contrary, adding the word Unit does make it undoubtably clear. So I will add it.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+        ctx.verifyChangeEventListenerCalledForActiveDoc(
+            times = 1,
+            expectedChangeEvent = ChangeEvent.changeEventForLocalInsert(
+                ctx.namespace,
+                ctx.testDocument,
+                true))
+        ctx.doSyncPass()
+        ctx.waitForEvent()
+        ctx.verifyChangeEventListenerCalledForActiveDoc(
+            times = 1,
```
adamchel: [overarching q] Since we're checking that the change event listener is called more than one time (once before doing the sync pass, and once after), does that mean end users will have `onEvent` called multiple times in their change event listeners for a single insert operation? that seems like incorrect behavior?

If that's not what's being tested here, can we clarify in the comments why we're waiting for the events again after doSyncPass? I see this behavior in a lot of the tests here

@edaniels would appreciate your insight on this

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+
+            if (shouldWaitForError) {
+                ctx.waitForError()
+            } else {
+                ctx.waitForEvent()
+            }
+
+            val expectedChangeEvent = if (shouldConflictBeResolvedByRemote)
+                ChangeEvent.changeEventForLocalDelete(ctx.namespace, ctx.testDocumentId, false)
+            else ChangeEvent.changeEventForLocalInsert(ctx.namespace, expectedDocument, true)
```
adamchel: [q] Why are we expecting a `changeEventForLocalInsert` instead of a `changeEventForLocalReplace`? This question also applies to everywhere else we expect an insert where we're really expecting a replace event.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+        setupPendingReplace(ctx, expectedDoc)
+
+        ctx.queueConsumableRemoteUpdateEvent()
+        ctx.doSyncPass()
+        assertEquals(expectedDoc, ctx.findTestDocumentFromLocalCollection())
+
+        // should be frozen since the operation type was unknown
+        ctx.queueConsumableRemoteUnknownEvent()
+        ctx.doSyncPass()
+        assertEquals(expectedDoc, ctx.findTestDocumentFromLocalCollection())
```
adamchel: [nit q] how are we actually testing that the document is frozen? shouldn't we try remote deleting it first before checking that it's still there?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+        ctx = harness.freshTestContext()
+        ctx.insertTestDocument()
+        ctx.queueConsumableRemoteUnknownEvent()
+        ctx.doSyncPass()
+        assertEquals(ctx.testDocument, ctx.findTestDocumentFromLocalCollection())
+
+        // should be frozen since the operation type was unknown
+        ctx.queueConsumableRemoteUpdateEvent()
+        ctx.doSyncPass()
+        assertEquals(ctx.testDocument, ctx.findTestDocumentFromLocalCollection())
```
adamchel: [nit q] same q as before, how do we know this is frozen? don't we need to try deleting first?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+        assertEquals(
+            ctx.testDocument,
+            ctx.findTestDocumentFromLocalCollection())
 
-        // set the doc to frozen and reload the configs
-        dataSynchronizer.getSynchronizedDocuments(namespace).forEach {
-            it.isFrozen = true
-        }
-        dataSynchronizer.reloadConfig()
+        // 2: Local wins
```
adamchel: [nit] where is `1.`?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+        ctx.deleteTestDocument()
+        ctx.waitForEvent()
+
+        ctx.verifyChangeEventListenerCalledForActiveDoc(1, expectedChangeEvent = expectedEvent)
+        val expectedException = StitchServiceException("bad", StitchServiceErrorCode.UNKNOWN)
+        ctx.mockDeleteException(expectedException)
+
+        ctx.doSyncPass()
+        ctx.waitForError()
+        // verify we are inserting!
```
adamchel: [nit] copy-pasted comment?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1,190 +1,882 @@
...
+
+    @Test
+    fun testDeleteOneById() {
+        val ctx = harness.freshTestContext()
+
+        // 0: Pre-checks
+        // assert this doc does not exist
+        assertNull(ctx.findTestDocumentFromLocalCollection())
+
+        // update the non-existent document...
```
adamchel: [nit] `update` -> `delete`

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -0,0 +1,1361 @@
...
+
+        // sync. assert that the conflict handler and
+        // change event listener have been called
+        streamAndSync()
+
+        changeEventListenerSemaphore.tryAcquire(3, TimeUnit.SECONDS)
```
adamchel: [major] another tryAcquire to `assertTrue` and change to 10 seconds

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncTestHarness.kt
```diff
@@ -0,0 +1,567 @@
...
+            Assert.assertEquals(expectedEvent.updateDescription.updatedFields, actualEvent.updateDescription.updatedFields)
+
+            Assert.assertEquals(expectedEvent.hasUncommittedWrites(), actualEvent.hasUncommittedWrites())
+        }
+
+        private fun newErrorListener(
+            emitErrorSemaphore: Semaphore? = null,
+            expectedDocumentId: BsonValue? = null
+        ): ErrorListener {
+            open class TestErrorListener : ErrorListener {
```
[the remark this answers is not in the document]
jsflax: At the moment, we don't need to interact with it. Might as well not muddle up what developers can access.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: Discussed offline.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: This event is specific to the circumstances that we are creating here. There are multiple ways to trigger a replace– this happens to be the way I've done it (based on the int tests).

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: I can add a case for this, though I believe I handle it with the previous update. There is also no "real" remote– we're testing this by queueing fake events as if they came from the remote.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: We don't need to try deleting– making sure the document doesn't update works as well.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: Adding now.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: Yes

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: 👍

jsflax:
@adamchel PTAL

PR #70, opened 2018-10-19
STITCH-1960 Conform to synchronization algorithm in specification

adamchel:
I did this by basically adding the spec to the code as comments. there were several places where I had to add missing functionality, and there’s at least one place where I’m not sure if we need to add more to the spec to make it complete.

I didn’t add any tests since I figured existing tests cover this, and some of the previously unimplemented features of the sync algorithm are more for optimization than correctness.

### DIFF
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
@@ -399,6 +399,7 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
         "t='%d': syncRemoteToLocal START",
         logicalT));
 
+    // 2. Run remote to local (R2L) sync routine
     for (final NamespaceSynchronizationConfig nsConfig : syncConfig) {
       final Map<BsonValue, ChangeEvent<BsonDocument>> remoteChangeEvents =
           instanceChangeStreamListener.getEventsForNamespace(nsConfig.getNamespace());
@@ -412,6 +413,7 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
         latestDocumentMap.put(latestDocument.get("_id"), latestDocument);
       }
 
+      // a. For each unprocessed change event
       for (final Map.Entry<BsonValue, ChangeEvent<BsonDocument>> eventEntry :
           remoteChangeEvents.entrySet()) {
         logger.info(String.format(
@@ -420,6 +422,7 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
             logicalT,
             eventEntry.getValue().getOperationType()));
 
+        // i. Find the corresponding local document.
         final CoreDocumentSynchronizationConfig docConfig =
             nsConfig.getSynchronizedDocument(eventEntry.getKey().asDocument().get("_id"));
 
@@ -433,6 +436,9 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
         syncRemoteChangeEventToLocal(nsConfig, docConfig, eventEntry.getValue());
       }
 
+      // For synchronized documents that had no unprocessed change event, but were marked as stale,
+      // synthesize a remote replace event to replace the local stale document with the latest
+      // remote copy.
       for (final BsonValue docId : unseenIds) {
         final CoreDocumentSynchronizationConfig docConfig =
             nsConfig.getSynchronizedDocument(docId);
@@ -456,6 +462,9 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
         }
       }
 
+      // For synchronized documents that had no unprocessed change event, and did not have a latest
+      // version when stale documents were queried, synthesize a remote delete event to delete
+      // the local document.
       unseenIds.removeAll(latestDocumentMap.keySet());
       for (final BsonValue unseenId : unseenIds) {
         final CoreDocumentSynchronizationConfig docConfig =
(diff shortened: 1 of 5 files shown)


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -654,6 +789,8 @@ private void syncLocalToRemote() {
           continue;
         }
 
+        // i. Retrieve the synthetic change event for this local document in the local config
```
jsflax: [nit] We may want to be careful about using the word `synthetic` here. Local events exist in a transient state, but are arguably real. Synthetic as term makes more sense when we are generating remote events based on conflict, and do not have actual data about what the remote change event was. I feel it's important that we do not conflate terminology here.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -678,208 +815,273 @@ private void syncLocalToRemote() {
...
-            // notification for it. There's nothing we can do about this.
-            try {
-              remoteColl.insertOne(withNewVersion(localChangeEvent.getFullDocument(), nextVersion));
-            } catch (final StitchServiceException ex) {
-              if (ex.getErrorCode() != StitchServiceErrorCode.MONGODB_ERROR
-                  || !ex.getMessage().contains("E11000")) {
-                this.emitError(docConfig, String.format(
+        // ii. Check if the internal remote change stream listener has an unprocessed event for
+        //     this document.
+        final ChangeEvent<BsonDocument> unprocessedRemoteEvent =
```
jsflax: Nice seeing this to spec.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListener.java
```diff
@@ -280,4 +282,21 @@ void storeNextEvent() {
...
     }
   }
+
+  /**
+   * If there is an unprocessed change event for a particular document ID, fetch it from the change
+   * stream listener without removing it.
+   *
+   * @return the latest unprocessed change event for the given document ID, or null if none exists.
+   */
+  public @Nullable ChangeEvent<BsonDocument> getUnprocessedEventForDocumentId(
```
jsflax: [minor?] Once this event is read, shouldn't it be considered processed? That was the logic behind clearing the map in `getEvents`– once they've been read, they have been processed. It seems like we should be removing them here.

edaniels: Agreed with Jason. This isn't a bug but it would cause us to keep doing work that's not needed.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -539,6 +547,8 @@ private void syncRemoteChangeEventToLocal(
       return;
     }
 
+    // TODO(QUESTION FOR REVIEW): This code is not in the spec, but if I don't include it, then
+    // testInsertThenSyncUpdateThenUpdate will fail. Should I add it to the spec?
```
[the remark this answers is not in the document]
edaniels: It's this in spec:
```
drop the event if the version counter of the remote event less than or equal to the version counter of the local document
```

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -420,6 +422,7 @@ private void syncRemoteToLocal() {
             logicalT,
             eventEntry.getValue().getOperationType()));
 
+        // i. Find the corresponding local document.
```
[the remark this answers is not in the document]
adamchel: Sounds good, but this means changing the spec as well since I copied directly from there.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
adamchel: But that's in a different place. The way the spec is written, this check is only made "if the document does not have local writes pending". Does this mean we should reorder the spec so that the check occurs before the "If the document does not have local writes pending" check?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
adamchel: Alright, removing the word synthetic, and updating the spec as well.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListener.java
adamchel: Ah, good catch, took me longer than it should have to wrap my head around this. I thought that because this is L2R we're not actually processing the event, and that we should wait for R2L to actually process it, but you're right that in any case, the event here is being effectively dropped or used to raise a conflict, so I'll fix this and clarify the comments here.

adamchel: Clarified the comments, and added a suggestion to the spec as well.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
edaniels: I suppose it should be changed in the spec. We do this check if we have changes or not.

adamchel:
@jsflax PTAL. 

also @edaniels I made relevant changes to the spec. if you make any changes to what I added, let me know so I can update the comments in the code

PR #71, opened 2018-10-23
Various sync fixes

edaniels:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -246,15 +246,18 @@ public void reloadConfig() {
   public void start() {
```
edaniels: `start` now creates a thread if one is not present and also stops the listeners. This is to prevent the general case of failed listeners and a sync thread that is not null when a network listener only reports that it is back online again, not offline. That's how the android one works. This doesn't cause any issues AFAIK.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -526,8 +529,23 @@ private void syncRemoteChangeEventToLocal(
         docConfig.getDocumentId(),
         remoteChangeEvent.getOperationType().toString()));
 
-    final DocumentVersionInfo currentRemoteVersionInfo = DocumentVersionInfo
-            .getRemoteVersionInfo(remoteChangeEvent.getFullDocument());
+    final DocumentVersionInfo currentRemoteVersionInfo;
```
edaniels: There's a bunch of these now. basically this guards us from pathological inputs.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1770,6 +1818,20 @@ void triggerListeningToNamespace(final MongoNamespace namespace) {
     }
   }
 
+  void triggerListeningToNamespace(
+      final MongoNamespace namespace, final BsonValue id, final boolean fromSync) {
+    syncLock.lock();
+    try {
+      if (instanceChangeStreamListener.isDocumentBeingWatched(namespace, id) == fromSync) {
```
edaniels: This variant is to make sure we only restart a listener if this is a new id being added.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1109,16 +1153,6 @@ private void syncLocalToRemote() {
                   committedEvent.getUpdateDescription(),
                   false));
 
-          if (nextVersion == null) {
```
edaniels: @adamchel I couldn't tell if this was really necessary. I could add this back and check that OP is not DELETE. Thoughts?

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoAdapter.java
```diff
@@ -28,17 +28,27 @@
...
+import java.util.Map;
+import java.util.Set;
+
 import org.bson.types.ObjectId;
 
 public class TodoAdapter extends RecyclerView.Adapter<TodoAdapter.TodoItemViewHolder> {
   private static final String TAG = TodoAdapter.class.getSimpleName();
   private final ItemUpdater itemUpdater;
   private List<TodoItem> todoItems;
+  private Map<ObjectId, Integer> todoItemIdsToIdx;
```
jsflax: [nit][q] Why the 'x' suffix?

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoAdapter.java
```diff
@@ -69,19 +79,68 @@ public int getItemCount() {
...
+    newTodoItems.addOnCompleteListener(task -> {
       if (!task.isSuccessful()) {
         Log.e(TAG, "failed to get items", task.getException());
+        return;
+      }
+      final Set<Integer> unseenIdxs = new HashSet<>(todoItemIdsToIdx.values());
+      final List<TodoItem> newItems = task.getResult();
+
+      // replace existing items and add new ones
+      for (final TodoItem newItem : newItems) {
```
jsflax: Interesting that this is a road a user might have to go down. We may want to consider (in the far future) simplifying processes like this for users.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
```diff
@@ -136,26 +128,70 @@ public void updateTask(final ObjectId itemId, final String currentTask) {
...
   }
 
   private class ItemUpdateListener implements ChangeEventListener<Document> {
     @Override
     public void onEvent(final BsonValue documentId, final ChangeEvent<Document> event) {
-      if (!event.hasUncommittedWrites()) {
-        todoAdapter.updateItems(getItems());
+      final Handler mainHandler = new Handler(Looper.getMainLooper());
+      if (event.getOperationType() == ChangeEvent.OperationType.DELETE) {
+        mainHandler.post(() ->
```
jsflax: [minor] Not running on the main thread is going to throw many users off. We should other call all event listeners on the main thread or heavily document this.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
```diff
@@ -306,30 +345,43 @@ private void showEditItemDialog(final ObjectId itemId, final String currentTask)
...
           touchList();
         })
-        .addOnFailureListener(e -> Log.e(TAG, "failed to insert log item", e));
+        .addOnFailureListener(e -> Log.e(TAG, "failed to insert todo item", e));
   }
 
   private void touchList() {
     lists.sync().updateOneById(
-        new BsonString(Stitch.getDefaultAppClient().getAuth().getUser().getId()),
+        new BsonString(userId),
```
jsflax: [q] Is it common to be using the raw Bson types instead of the Java-coerced ones? Seems like we should be using `String` instead of `BsonString` wherever we can. Probably a comment for GA though.

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/internal/SyncMongoClientIntTests.kt
```diff
@@ -928,6 +928,8 @@ class SyncMongoClientIntTests : BaseStitchAndroidIntTest() {
             { _, _ -> }
         )
 
+        waitForAllStreamsOpen()
```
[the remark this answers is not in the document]
jsflax: 🎉

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/internal/SyncMongoClientIntTests.kt
```diff
@@ -1166,13 +1168,18 @@ class SyncMongoClientIntTests : BaseStitchAndroidIntTest() {
         }
     }
 
+    private fun waitForAllStreamsOpen() {
+        val dataSync = (mongoClient as RemoteMongoClientImpl).dataSynchronizer
+        while (!dataSync.areAllStreamsOpen()) {
+            println("waiting for all streams to open before doing sync pass")
+            Thread.sleep(1000)
```
jsflax: [nit] Maybe add a fail case here if we wait too long.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Nice. What do you mean pathological inputs? Just not a phrase I'm familiar with.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Nice, I like this change.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoAdapter.java
edaniels: Index :)

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoAdapter.java
edaniels: I think this is just crappy application design on my part and not using android adapters well. I think Room improves this.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
edaniels: good point. So Tasks run their handlers on the main thread. It might be good for us to default to the main thread for this. Do you know if running event based actions tends to be on the main thread? I think the answer is yes given the event based system for Android

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
edaniels: Well we require updateOneById to accept a BsonValue. We could make it accept an Object and hope for the best that it serializes. I'm open to changing it. @adamchel ? I just went with type safety initially.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
edaniels: I just mean inputs designed to go against how this code is supposed to function. So pathological in the mathematical sense. This is also for malicious input of a user trying to break sync.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
adamchel: Yeah I agree we should run on the main thread, especially if we're gonna do the same on iOS.

android/examples/todo-sync/src/main/java/com/mongodb/stitch/android/examples/todosync/TodoListActivity.java
adamchel: Hmm, I think type safety is important, let's see how people feel about this during beta.

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -329,43 +348,53 @@ private Response handleAuthFailure(final StitchServiceException ex, final Stitch
...
-      throw new StitchRequestException(e, StitchRequestErrorCode.DECODING_ERROR);
+      // retry
+      refreshAccessToken();
+    } finally {
+      authLock.unlock();
     }
+  }
 
+  void refreshAccessToken() {
+    authLock.lock();
```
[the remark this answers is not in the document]
edaniels: This used to be synchronized so it works exactly the same with a ReentrantLock (it's recursive)

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1823,16 +1885,7 @@ public Object call() {
                 "emitEvent ns=%s documentId=%s emit exception: %s",
                 event.getNamespace(),
                 documentId,
-                ex));
-            emitError(
```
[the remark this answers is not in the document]
edaniels: This isn't an error that should freeze something. It's a user level exception they should catch.

PR #73, opened 2018-10-28
STITCH-2133 Add ability to unfreeze document via sync

jsflax:
The main addition of this ticket is the added method `resumeSyncForDocument`. A major secondary change is that the language for "frozen/thawed" has been switched to "paused/resumed". This is the documentation for this method:

```
/**
   * A document that is paused no longer has remote updates applied to it.
   * Any local updates to this document cause it to be resumed. An example of pausing a document
   * is when a conflict is being resolved for that document and the handler throws an exception.
   *
   * @param documentId the id of the document to resume syncing
   * @return true if successfully resumed, false if the document
   *         could not be found or there was an error resuming
*/
boolean resumeSyncForDocument(@Nonnull final BsonValue documentId);
```

This method is proxied upwards from `DataSynchronizer` => `CoreSync` => `Sync`.

The tests assert that, in the event of an irrecoverable error, a document is paused, and that document can be resumed via this method. The method should return the appropriate boolean value if the resume was successful. Operations should then be able to be sync'd on said document.

### DIFF
--- a/android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
+++ b/android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
@@ -84,6 +84,14 @@ public interface Sync<DocumentT> {
   Set<BsonValue> getSyncedIds();
 
   /**
+   * Return the set of synchronized document _ids in a namespace
+   * that have been paused due to an irrecoverable error.
+   *
+   * @return the set of paused document _ids in a namespace
+   */
+  Set<BsonValue> getPausedDocumentIds();
+
+  /**
    * A document that is paused no longer has remote updates applied to it.
    * Any local updates to this document cause it to be resumed. An example of pausing a document
    * is when a conflict is being resolved for that document and the handler throws an exception.
--- a/android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
+++ b/android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
@@ -84,6 +84,15 @@ public interface Sync<DocumentT> {
   Set<BsonValue> getSyncedIds();
 
   /**
+   * Unfreeze a document.
+   *
+   * @param documentId the id of the document to unfreeze
+   * @return true if successfully unfrozen, false if the document
+   *         could not be found or there was an error unfreezing
+   */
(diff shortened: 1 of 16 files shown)


android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
```diff
@@ -83,6 +83,17 @@ void configure(@NonNull final ConflictHandler<DocumentT> conflictHandler,
...
+  /**
+   * A document that is paused no longer has remote updates applied to it.
+   * Any local updates to this document cause it to be resumed. An example of pausing a document
+   * is when a conflict is being resolved for that document and the handler throws an exception.
+   *
+   * @param documentId the id of the document to resume syncing
+   * @return true if successfully resumed, false if the document
+   *         could not be found or there was an error resuming
+   */
+  boolean resumeSyncForDocument(@NonNull final BsonValue documentId);
```
edaniels: This is good. In addition to this, we need a way to get the Set of all documents that are frozen

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1526,6 +1526,37 @@ public void desyncDocumentFromRemote(
...
+    final NamespaceSynchronizationConfig namespaceSynchronizationConfig;
+    final CoreDocumentSynchronizationConfig config;
+
+    if ((namespaceSynchronizationConfig = syncConfig.getNamespaceConfig(namespace)) == null
+        || (config = namespaceSynchronizationConfig.getSynchronizedDocument(documentId)) == null) {
+      return false;
+    }
+
+    config.setPaused(false);
+    return !config.isPaused();
```
edaniels: won't this always be true?

jsflax: Thought there's a small chance of a race here– document is resumed and in that instant an error occurs again. Returning !isPaused seemed like a sure way to return the state of the doc. I could leave a TODO for your threading ticket. Or I could be being overly careful here.

edaniels: I feel like it's fine to return true. The same thing can happen even right after it returns to you.

PR #74, opened 2018-10-31
STITCH-1967 Add full CRUD within .sync()

jsflax:
_Ticket Description:_
_This should implement the same CRUD methods that the non sync service has. All of these methods should only operate on the synced, offline documents._

_For replacements, they should use the same set and unset event optimization_

1. Add `insertMany`, `updateMany`, `deleteMany`, `aggregate`, and `count` to the DataSynchronizer. This is to be in parity with `RemoteMongoCollection`.
    a. For `deleteMany` and `updateMany`, the documents need to be fetched before the delete to 
        retrieve the ids, due to the way canonical filters work.
    b. For `updateMany`, this will match potential unmodified documents. Ignore those.
    c. For `updateMany`, it may be an upsert. If it is an upsert, filter the result using the upsertedId
        and emit an insert.
2. Replace `findOneById`, `updateOneById` and `deleteOneById` with `find`, `updateOne` and `deleteOne`, respectively.
    a. For `deleteOne`, the document needs to be fetched before the delete to retrieve the id, due to 
        the way canonical filters work.
    b. For `updateOne`, it may be an upsert. If it is an upsert, filter the result using the upsertedId
        and emit an insert.
3. Proxy each method upwards through `CoreSync`.
4. Add `AggregateOperation`, `CoreSyncAggregateIterable`, `CountOperation`, `InsertManyOperation`, and `DeleteManyOperation` classes to be instantiated in SyncOperations to maintain parity with `Operations` (remote).
5. Add `SyncCountOptions`, `SyncDeleteOptions`, `SyncUpdateOptions`, `SyncDeleteResult`, `SyncInsertManyResult`, `SyncInsertOneResult`, and `SyncUpdateResult` to both maintain parity with the Remote*Options classes, and to create a divergence between our options and the local mongo client library.
6. Proxy `CoreSync` operations upwards through `Sync` (server) and `Sync` (Android).

Unit testing to test the new logic was critical here, especially for the bulk operations. Integration tests were limited to ensuring that each new operation operates as intended in an integrated environment. Since no new Sync cases were created, adding new conflicted/failing integration tests seemed unnecessarily cumbersome.


jsflax:
@edaniels @adamchel PTAL

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncUnitTestHarness.kt
```diff
@@ -354,8 +366,16 @@ class SyncUnitTestHarness : Closeable {
                 bsonDocumentCodec)
         }
 
-        override fun waitForEvent() {
-            assertTrue(eventSemaphore?.tryAcquire(10, TimeUnit.SECONDS) ?: true)
+        override fun waitForEvents(amount: Int) {
+            waitLock.lock()
+            changeEventListener.totalEventsToAccumulate = amount
```
edaniels: [q] Why is this here?

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncUnitTestHarness.kt
```diff
@@ -220,20 +207,45 @@ class SyncUnitTestHarness : Closeable {
...
         }
     }
 
     @Suppress("UNCHECKED_CAST")
     private class DataSynchronizerTestContextImpl(shouldPreconfigure: Boolean = true) : DataSynchronizerTestContext {
+        open class TestChangeEventListener(
+            private val expectedEvent: ChangeEvent<BsonDocument>?,
+            var emitEventSemaphore: Semaphore?
+        ) : ChangeEventListener<BsonDocument> {
+            val eventAccumulator = mutableListOf<ChangeEvent<BsonDocument>>()
```
edaniels: [q] Why was this added?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2056,13 +2299,13 @@ private static BsonDocument getDocumentIdFilter(final BsonValue documentId) {
   /**
    * Adds and returns a document with a new version to the given document.
    *
-   * @param document the document to attach a new version to.
+   * @param document   the document to attach a new version to.
```
edaniels: Is checkstlye not catching any of these issues?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2056,13 +2299,13 @@ private static BsonDocument getDocumentIdFilter(final BsonValue documentId) {
...
    * @param newVersion the version to attach to the document
    * @return a document with a new version to the given document.
    */
   private static BsonDocument withNewVersion(
-          final BsonDocument document,
-          final BsonDocument newVersion
+      final BsonDocument document,
```
edaniels: Why did this change?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1543,20 +1548,21 @@ public void desyncDocumentFromRemote(
...
+    getLocalCollection(namespace).deleteOne(getDocumentIdFilter(documentId));
     triggerListeningToNamespace(namespace);
   }
 
   /**
    * A document that is paused no longer has remote updates applied to it.
    * Any local updates to this document cause it to be resumed. An example of pausing a document
    * is when a conflict is being resolved for that document and the handler throws an exception.
-   *
+   * <p>
```
edaniels: Stray?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1634,71 +1674,221 @@ boolean resumeSyncForDocument(
...
+    final ChangeEvent<BsonDocument> event;
+    final CoreDocumentSynchronizationConfig config;
+    final BsonValue documentId = BsonUtils.getDocumentId(documentAfterUpdate);
+
+    // if there was no document prior and this was an upsert,
+    // treat this as an insert.
+    // else this is an update
+    if (documentBeforeUpdate == null && updateOptions.isUpsert()) {
+      config = syncConfig.addSynchronizedDocument(namespace, documentId);
+      triggerListeningToNamespace(namespace);
```
edaniels: Curious, do you have a test that checks that if you were to remove this line, it would fail?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1634,71 +1674,221 @@ boolean resumeSyncForDocument(
...
+      updatedFilter = new BsonDocument("_id", new BsonDocument("$in", ids));
+    }
+
+    this.getLocalCollection(namespace).find(updatedFilter).forEach(new Block<BsonDocument>() {
+      @Override
+      public void apply(@NonNull final BsonDocument afterDocument) {
+        final BsonDocument beforeDocument;
+        final BsonValue documentId = BsonUtils.getDocumentId(afterDocument);
+
+        if ((beforeDocument = idToBeforeDocumentMap.get(documentId)) == null
```
edaniels: Comment here

edaniels: [nit] Assigning to beforeDocument like this feels kind of odd as opposed to just doing it right before the if

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncUnitTestHarness.kt
jsflax: This is to give the harness the ability to wait for multiple events. The accumulator actually does nothing– could just be a counter– as the actual events are later accumulated via capturing the values from the spy.

This is being used to test bulk writes, as multiple change events are emitted from a single method call.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncUnitTestHarness.kt
jsflax: See answer above.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Checkstyle did not. The DataSynchronizer doesn't seem to actually be formatted, as far as I can tell, and my autoformatter "went to town" on the class. We should probably file a ticket (or just make a patch) to reformat it.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: See above.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: See above. Though technically, this is also fine.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: I don't believe so– I can add a line in our new upsert tests that ensures that the stream is opened on the namespace.

jsflax: Added to `testUpsertOne` and `testUpsertMany` in `DataSynchronizerUnitTests` 👍

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: 👍

jsflax: 👍

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: 👍

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/ProxySyncMethods.kt
```diff
@@ -55,43 +58,89 @@ interface ProxySyncMethods {
...
+    /**
+     * Inserts one or more documents. Begin synchronizing on the documents' ids.
+     *
+     * @param documents the documents to insert
+     * @return the result of the insert many operation
+     */
+    fun insertManyAndSync(documents: List<Document>): SyncInsertManyResult
 
     /**
      * Deletes a single document by the given id. It is first searched for in the local synchronized
```
adamchel: [nit] this comment looks out of date. honestly this file probably doesn't need doc comments for the CRUD methods, this is pretty internal

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1634,71 +1674,229 @@ boolean resumeSyncForDocument(
...
-      final BsonDocument update
-  ) {
-    // TODO: lock down id
-    final CoreDocumentSynchronizationConfig config =
-        syncConfig.getSynchronizedDocument(namespace, documentId);
-    if (config == null) {
-      return UpdateResult.acknowledged(0, 0L, null);
+  void insertManyAndSync(final MongoNamespace namespace,
+                         final List<BsonDocument> documents) {
+    getLocalCollection(namespace).insertMany(documents);
```
adamchel: [alternatively] instead of generating IDs client-side, and depending on `BsonUtils.getDocumentId` not failing we could look at the result of the local `insertMany` (and `insertOne` in `insertOneAndSync`), and use the inserted ids there to update the doc config and emit the event

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1634,71 +1674,229 @@ boolean resumeSyncForDocument(
...
+      return UpdateResult.acknowledged(0, 0L, null);
+    }
 
-    if (result == null) {
+    // find and update the single document, returning the document post-update
+    final BsonDocument documentAfterUpdate = localCollection.findOneAndUpdate(
+        filter,
+        update,
+        new FindOneAndUpdateOptions()
+            .collation(updateOptions.getCollation())
```
adamchel: [minor] collation, bypassDocumentValidation, and arrayFilters are all unsupported by the remote MongoDB service so this might result in conflicts between local and remote where a document that is able to inserted locally is not able to be inserted remotely. we could either not support those options at all, or just do this with the assumption that we will eventually support the full mongodb surface area

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -832,7 +943,212 @@ class DataSynchronizerUnitTests {
     }
 
     @Test
-    fun testDeleteOneById() {
+    fun testUpsertOne() {
```
adamchel: [nit] one other test that might be worth adding is to do an update where upsert is specified, but there are documents that match, and we want to verify that no insert occurs

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -1429,6 +1346,188 @@ class SyncIntTestProxy(private val syncTestRunner: SyncIntTestRunner) {
...
+        localFound = coll.find(Document("fish", Document("\$exists", true)))
+
+        assertEquals(2, localFound.toList().size)
+        assertEquals(2, remoteFound.toList().size)
+        localFound.forEach { assertEquals(listOf("trout", "mackerel", "cod", "hake"), it!!["fish"]) }
+        remoteFound.forEach { assertEquals(listOf("trout", "mackerel", "cod", "hake"), it!!["fish"]) }
+    }
+
+    @Test
+    fun testDeleteManyNoConflicts() {
```
adamchel: [nit] might be worth throwing in an upsert test here

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreSyncUnitTests.kt
```diff
@@ -117,49 +142,103 @@ class CoreSyncUnitTests {
...
+
+        verify(syncOperations, times(1)).aggregate(eq(pipeline), eq(BsonDocument::class.java))
+
+        verify(ctx.dataSynchronizer, times(1)).aggregate(
+            eq(ctx.namespace), eq(pipeline), eq(BsonDocument::class.java))
     }
 
     @Test
-    fun testUpdateOneById() {
+    fun testUpdateOne() {
```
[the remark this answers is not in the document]
jsflax: These tests are already in the PR.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: There's no actual way for these to get passed downstream, if you look through the proxies– we do not actually support the options. But we seem to operate within the DataSynchronizer on localmdb without considering Remote.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
jsflax: I believe that's covered here– but I can add another check to make it clearer.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/AggregateOperation.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+import org.bson.conversions.Bson;
+
+class AggregateOperation<T> implements Operation<Collection<T>> {
```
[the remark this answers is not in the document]
jsflax: Discussed offline. Because these are package private, we are going to keep these as is.

`SyncFindOperation` will be whittled down to `FindOperation` as well.

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/internal/SyncMongoClientIntTests.kt
```diff
@@ -71,20 +76,36 @@ class SyncMongoClientIntTests : BaseStitchAndroidIntTest(), SyncIntTestRunner {
             sync.syncOne(id)
         }
 
-        override fun insertOneAndSync(document: Document): RemoteInsertOneResult {
+        override fun count(filter: Bson): Long {
+            return Tasks.await(sync.count())
```
dkaminsky: Even if not used currently, expected behavior would be for this to propagate the filter.
```suggestion
            return Tasks.await(sync.count(filter))
```

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncOperations.java
```diff
@@ -128,16 +154,46 @@ public InsertOneAndSyncOperation insertOneAndSync(
...
+      if (getCodec(codecRegistry, documentClass) instanceof CollectibleCodec) {
+        bsonDocuments.add(
+            documentToBsonDocument(
+                ((CollectibleCodec<DocumentT>) getCodec(codecRegistry, documentClass))
+                    .generateIdIfAbsentFromDocument(document),
+                codecRegistry
+            )
+        );
+      } else {
+        bsonDocuments.add(documentToBsonDocument(document, codecRegistry));
```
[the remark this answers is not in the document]
jsflax: I'm a bit confused where this is coming from. Is this code we have ever had? I thought embedded adds the ID.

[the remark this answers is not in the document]
adamchel: embedded does add the ID, but you're pulling the ID from the document before it's inserted. see my `[alternative]` comment for a solution that leverages the fact that embedded adds the id

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Should note that there are no results from local inserts.

jsflax: The local client does not return a result.

jsflax:
@adamchel PTAL

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreSyncUnitTests.kt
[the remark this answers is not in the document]
adamchel: yep, my bad, made this comment before I got to that part

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
adamchel: my mistake, I assumed it returned a an insertedId like Stitch

PR #77, opened 2018-11-03
STITCH-1972 Prevent modification of version field

adamchel:
This is a relatively small PR, but it was pretty tricky to get right. I tried to think of all the edge cases I could but I’d definitely appreciate careful eyes on this one.

I implemented the version filtering in such a way that it only has to happen when writing to the local collection. Any time we insert or update documents locally, or replace documents locally with a resolved doc, I filter out the version field (and if it’s already committed in the case of a local update, I revert that part of the update). With Eric’s new namespace locking, reads don’t have to worry about ever seeing a document with a version, since in an unlocked and recovered state, there shouldn’t ever be a document in the collection with a version.

Also for the sake of performance, I opted in some places to go with the pattern of mutating a method’s argument, but we said that was not best practice earlier, so I’m open to doing it another way.

Testing:
* I added an integration test specifically testing pathological cases where we try to update the version of a local document
* I updated all the integration tests to get rid of now unnecessary `withoutSyncVersion` calls
* I added checks in `testSyncInBothDirections` to ensure that any documents that exist in the local collection at the end of the test do not have any version information.

I didn’t add or edit any unit tests, which I think is fine, since this shouldn’t have really changed any functionality, and it feels a little redundant on top of the integration tests, but I can add them if you guys think it’s necessary.

Drive-bys:
* Fixed a bug where updates were not setting a version if the update was only `$unset`-ing fields
* Fixed a behavior (bug?) where updates that didn’t make any changes to a document were still incrementing the version counter. 
    * To do this, I had to modify the change event coalescence algorithm in the spec to consider inserts after deletes as replaces rather than updates. This is because it’s not possible to always produce an accurate update description without knowing the original undeleted document. To determine whether or not an update will make changes to a document (and thus whether the version should be incremented), we use the update description.

### DIFF
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
@@ -387,13 +387,13 @@ class CoreDocumentSynchronizationConfig {
         break;
       case DELETE:
         switch (newestChangeEvent.getOperationType()) {
-          // Coalesce inserts to updates since we believe at some point a document existed remotely
-          // and that this insert should really be an update if we are still in an uncommitted
-          // state.
+          // Coalesce inserts to replaces since we believe at some point a document existed
+          // remotely and that this insert should really be an replace if we are still in an
+          // uncommitted state.
           case INSERT:
             return new ChangeEvent<>(
                 newestChangeEvent.getId(),
-                ChangeEvent.OperationType.UPDATE,
+                ChangeEvent.OperationType.REPLACE,
                 newestChangeEvent.getFullDocument(),
                 newestChangeEvent.getNamespace(),
                 newestChangeEvent.getDocumentKey(),
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
@@ -1798,16 +1798,16 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
    */
   void insertOne(final MongoNamespace namespace, final BsonDocument document) {
     // Remove forbidden fields from the document before inserting it into the local collection.
-    sanitizeDocument(document);
+    final BsonDocument docForStorage = sanitizeDocument(document);
 
     final Lock lock = this.syncConfig.getNamespaceConfig(namespace).getLock().writeLock();
     lock.lock();
     final ChangeEvent<BsonDocument> event;
     final BsonValue documentId;
     try {
-      getLocalCollection(namespace).insertOne(document);
-      documentId = BsonUtils.getDocumentId(document);
-      event = changeEventForLocalInsert(namespace, document, true);
+      getLocalCollection(namespace).insertOne(docForStorage);
+      documentId = BsonUtils.getDocumentId(docForStorage);
+      event = changeEventForLocalInsert(namespace, docForStorage, true);
       final CoreDocumentSynchronizationConfig config = syncConfig.addSynchronizedDocument(
           namespace,
           documentId
@@ -1826,10 +1826,12 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
    * @param documents the documents to insert
    */
   void insertMany(final MongoNamespace namespace,
-                         final List<BsonDocument> documents) {
+                  final List<BsonDocument> documents) {
     // Remove forbidden fields from the documents before inserting them into the local collection.
+    final List<BsonDocument> docsForStorage = new ArrayList<>(documents.size());
+
     for (final BsonDocument document : documents) {
-      sanitizeDocument(document);
+      docsForStorage.add(sanitizeDocument(document));
     }
 
     final Lock lock =
@@ -1837,8 +1839,8 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
     lock.lock();
     final List<ChangeEvent<BsonDocument>> eventsToEmit = new ArrayList<>();
     try {
-      getLocalCollection(namespace).insertMany(documents);
-      for (final BsonDocument document : documents) {
+      getLocalCollection(namespace).insertMany(docsForStorage);
+      for (final BsonDocument document : docsForStorage) {
         final BsonValue documentId = BsonUtils.getDocumentId(document);
         final ChangeEvent<BsonDocument> event =
             changeEventForLocalInsert(namespace, document, true);
@@ -1909,7 +1911,7 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
       }
 
       // find and update the single document, returning the document post-update
-      final BsonDocument documentAfterUpdate = localCollection.findOneAndUpdate(
+      final BsonDocument unsanitizedDocumentAfterUpdate = localCollection.findOneAndUpdate(
           filter,
           update,
           new FindOneAndUpdateOptions()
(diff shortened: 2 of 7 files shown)


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1082,29 +1083,38 @@ private void syncLocalToRemote() {
...
+                if (!sets.isEmpty() || !unsets.isEmpty()) {
+                  // set the document version if any changes were made by this update
+                  sets.put(DOCUMENT_VERSION_FIELD, nextVersion);
+                  translatedUpdate.put("$set", sets);
+                }
 
                 final RemoteUpdateResult result;
                 try {
+                  if (translatedUpdate.isEmpty()) {
+                    // if the translated update is empty, then this update is a noop, and we
```
edaniels: [q] Should we log here? Also, maybe we should just check if it's a noop earlier on so that getting this branch is technically impossible.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2564,6 +2602,71 @@ private static BsonDocument getDocumentIdFilter(final BsonValue documentId) {
...
+    if (document == null) {
+      return null;
+    }
+
+    final BsonDocument filteredDoc = document.clone();
+    filteredDoc.remove(DOCUMENT_VERSION_FIELD);
+    return filteredDoc;
+  }
+
+  // (TODO) NOTE FOR REVIEWER: the reason I chose to make these functions mutate their args rather
```
edaniels: I think that's fine.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
adamchel: added the check earlier in the `UPDATE` and added a log statement, but this branch was definitely possible to begin with. Any update that matches documents but doesn't end up changing them locally will produce a noop event until we modify the update functions to not emit events for noop updates (and also properly return a modifiedCount that reflects these noop updates).

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -387,13 +387,13 @@ public boolean hasCommittedVersion(final DocumentVersionInfo versionInfo) {
...
-          // and that this insert should really be an update if we are still in an uncommitted
-          // state.
+          // Coalesce inserts to replaces since we believe at some point a document existed
+          // remotely and that this insert should really be an replace if we are still in an
+          // uncommitted state.
           case INSERT:
             return new ChangeEvent<>(
                 newestChangeEvent.getId(),
-                ChangeEvent.OperationType.UPDATE,
+                ChangeEvent.OperationType.REPLACE,
```
dkaminsky: Still new at this, but doesn't this have an impact for rules and triggers?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2564,6 +2617,67 @@ private static BsonDocument getDocumentIdFilter(final BsonValue documentId) {
...
+   * @param document The document from which to create a clone without forbidden fields.
+   * @return a clone of the given document without forbidden fields
+   */
+  static BsonDocument withoutForbiddenFields(final BsonDocument document) {
+    if (document == null) {
+      return null;
+    }
+
+    final BsonDocument filteredDoc = document.clone();
+    filteredDoc.remove(DOCUMENT_VERSION_FIELD);
```
dkaminsky: should this be driven by a Collection of constants so that we can easily add more things here should we need to without adding individual calls to "remove"?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1801,6 +1827,11 @@ void insertOne(final MongoNamespace namespace, final BsonDocument document) {
    */
   void insertMany(final MongoNamespace namespace,
                          final List<BsonDocument> documents) {
+    // Remove forbidden fields from the documents before inserting them into the local collection.
+    for (final BsonDocument document : documents) {
+      sanitizeDocument(document);
```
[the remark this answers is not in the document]
adamchel: mutating a user's data structures would only ever happen when a user is intentionally trying to use the forbidden field, which we will explicitly document not to do. I opted for the mutation so as not to take the performance hit of copying the entire document, which could be up to 16MB.

Though I guess I could have `sanitizeDocument` return a document, and do two passes on the document to avoid mutating: one to check if the forbidden field exists, and another to copy it if the forbidden field is present. In almost every case it will just return the same document reference rather than a copy.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1773,8 +1797,10 @@ long count(final MongoNamespace namespace, final Bson filter, final CountOptions
    * @param document  the document to insert.
    */
   void insertOne(final MongoNamespace namespace, final BsonDocument document) {
-    final Lock lock =
-        this.syncConfig.getNamespaceConfig(namespace).getLock().writeLock();
+    // Remove forbidden fields from the document before inserting it into the local collection.
+    sanitizeDocument(document);
```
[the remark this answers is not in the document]
adamchel: see above

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2090,6 +2129,11 @@ private void updateOrUpsertOneFromResolution(
       if (documentBeforeUpdate != null) {
         undoCollection.insertOne(documentBeforeUpdate);
       }
+
+      // Remove forbidden fields from the resolved document before it will updated/upserted in the
+      // local collection.
+      sanitizeDocument(document);
```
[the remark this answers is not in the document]
adamchel: see above

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
edaniels: Yes but I don't think there's a good solution for this if the user thinks it's really an insert. We can't hold on to enough date for this case.

adamchel: per offline discussion, the answer is that nothing really changed with this particular update since the server was receiving the same events in the end.

your concerns are valid, but they're more because of the limitations of change streams than they are because of the way we've implemented sync

adamchel:
@dkaminsky PTAL. just implemented the strategy of not mutating the document, and cloning only when necessary

PR #78, opened 2018-11-14
STITCH-2185 Fix various issues with sync recovery logic, and add unit tests

adamchel:
This PR supplements our sync unit tests with tests that do the following
* Verifies that undo documents are never leaked when write operations complete successfully.
* Verifies that several failure scenarios are recovered properly.
* Verifies that pathological inputs unrelated to synchronized documents are not stored in the local synchronized collection.

In writing these tests, I found various bugs in the recovery logic that should now be fixed. Some documents were getting leaked, and rollbacks sometimes occurred when they shouldn't have.

Also did a drive-by change to expose the min age and max age in a user profile to be Strings rather than ints. This follows the recent change to the iOS SDK, and protects the SDK from min_age and max_age String inputs that don't represent integer values.

### DIFF
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
@@ -256,6 +256,9 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
               localCollection.deleteOne(filter);
               break;
             case UNKNOWN:
+              throw new IllegalStateException(
+                      "there should not be a pending write with an unknown event type"
+              );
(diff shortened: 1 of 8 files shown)


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -222,11 +225,62 @@ private void recoverNamespace(final NamespaceSynchronizationConfig nsConfig) {
...
+                      filter,
+                      pendingWrite.getFullDocument(),
+                      new FindOneAndReplaceOptions().upsert(true)
+              );
+              break;
+            case DELETE:
+              localCollection.deleteOne(filter);
+              break;
+            default:
+              // TODO(question for review): should I throw IllegalStateException here? There should
```
edaniels: I think it's totally okay to throw here.

PR #80, opened 2018-11-19
STITCH-2008 - code complete, no unit tests yet - creates an 'app clie…

dkaminsky:
…nt' subclass of the request client that uses a client app ID to figure out the localized base URL for requests

Submitting now so Eric has a chance to see it today as it will be a template for other SDKs. Continuing to work on unit tests during review. All existing tests are passing.

The purpose of this ticket is to ensure that calls to the client v2 API are directed to the proper URL, which is the configured "base URL" for "globally deployed apps" but is a localized URL for "locally deployed apps." These URLs are controlled in the server configuration and exposed to the client via the endpoint (created in STITCH-2001) that returns application metadata based on the client app ID.

This implementation wraps the core request client in an app request client that is initialized with the relevant client app ID and lazily (ie. on the first request) fills in the application metadata for that ID by requesting it from the server directly using the underlying transport. As a result, that first user request and all subsequent requests will be directed to the proper URL based on the application's deployment model.

### DIFF
--- a/core/admin-client/src/main/java/com/mongodb/stitch/core/admin/StitchAdminClient.kt
+++ b/core/admin-client/src/main/java/com/mongodb/stitch/core/admin/StitchAdminClient.kt
@@ -2,7 +2,10 @@ package com.mongodb.stitch.core.admin
 
 import com.mongodb.stitch.core.auth.StitchCredential
 import com.mongodb.stitch.core.internal.common.MemoryStorage
-import com.mongodb.stitch.core.internal.net.*
+import com.mongodb.stitch.core.internal.net.Method
+import com.mongodb.stitch.core.internal.net.OkHttpTransport
+import com.mongodb.stitch.core.internal.net.StitchAuthRequest
+import com.mongodb.stitch.core.internal.net.StitchRequestClientImpl
 
 class StitchAdminClient private constructor(
     private val adminAuth: StitchAdminAuth,


core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchRequest.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Arrays;
```
edaniels: This file can be reverted for the sake of minimizing code change. We can driveby format later.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * @param stitchReq the request to perform.
+   * @return a {@link Response} to the request.
+   */
+  @Override
+  public Response doRequest(final StitchRequest stitchReq) {
+    if (appMetadata == null) {
+      init(clientAppId);
+    }
+    // init throws if no location is available, so this should be safe
```
edaniels: [nit] I think you can remove these types of comments and assume init will get you appMetadata. In fact, could rename `init` to `initAppMetadata`.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+
+    try {
+      return transport.stream(buildRequest(stitchReq, appMetadata.hostname));
+    } catch (Exception e) {
+      throw new StitchRequestException(e, StitchRequestErrorCode.TRANSPORT_ERROR);
+    }
+  }
+
+  private synchronized void init(final String clientAppId) {
+    // double-check, in case two threads hit this point simultaneously
```
edaniels: Can be avoided by adding synchronized up front.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+    } else {
+      throw new StitchRequestException(
+          String.format(BOOTSTRAP_ERROR_MESSAGE_INVALID_HOSTNAME,
+              appMetadata == null ? "null" : appMetadata),
+          StitchRequestErrorCode.BOOTSTRAP_ERROR);
+    }
+  }
+
+  @SuppressWarnings("unused") // Jackson uses reflection
+  static class AppMetadata {
```
edaniels: [style] Move this out to its own class/file outside of here.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+    public String getHostname() {
+      return hostname;
+    }
+
+    public void setHostname(final String hostname) {
+      this.hostname = hostname;
+    }
+
+    @Override
+    public String toString() {
```
edaniels: [q] Is this used/needed?

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+   * @return a {@link Response} to the request.
+   */
+  @Override
+  public Response doRequest(final StitchRequest stitchReq) {
+    if (appMetadata == null) {
+      init(clientAppId);
+    }
+    // init throws if no location is available, so this should be safe
+
+    final Response response;
```
edaniels: [minor] Given this class extends `StitchRequestClient`, we should try to reuse its methods as much as possible to avoid the risk of missing out on logic that it has and we forget to implement here (and preserve bug compatibility), if it exists. To that end, I would add overloaded methods to `StitchRequestClient` for `doRequest` and `doStreamRequest` that take the URL discovered here. That way those methods can be very short stubs that end up calling the original `doRequest`/`doStreamRequest`.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+      return;
+    }
+
+    final StitchAppRoutes routes = new StitchAppRoutes(clientAppId);
+
+    final StitchRequest bootstrapStitchRequest = new StitchRequest.Builder()
+        .withMethod(Method.GET)
+        .withPath(routes.getServiceRoutes().getLocationRoute())
+        .build();
+    final Request bootstrapRequest = buildRequest(bootstrapStitchRequest, baseUrl);
```
edaniels: [minor] For the code here and below, you should just be able to reuse super's `doRequest` and then decode it like you are at the bottom here. That would avoid some duplicate code.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    inspectResponse(response);
+
+    final AppMetadata responseMetadata;
+    try {
+      responseMetadata = StitchObjectMapper.getInstance()
+          .readValue(IoUtils.readAllToString(response.getBody()), AppMetadata.class);
+
+    } catch (Exception e) {
```
edaniels: [style] Downgrade this to a `final IOException e`

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,180 @@
...
+      throw new StitchRequestException(e, StitchRequestErrorCode.DECODING_ERROR);
+    }
+
+    if (responseMetadata != null && responseMetadata.hostname != null) {
```
edaniels: [nit] In addition to checking the hostname is not null, I'd check that it's also non empty.

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchRequest.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: discussed offline

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
dkaminsky: done

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/StitchAppRequestClient.java
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      responseMetadata = StitchObjectMapper.getInstance()
+          .readValue(IoUtils.readAllToString(response.getBody()), AppMetadata.class);
+
+    } catch (IOException e) {
+      throw new StitchRequestException(e, StitchRequestErrorCode.DECODING_ERROR);
+    }
+
+    if (responseMetadata != null && responseMetadata.hostname != null
+        && !"".equals(responseMetadata.hostname.trim())) {
```
adamchel: [nit] `!responseMetadata.hostname.trim().isEmpty()` instead of `"".equals(...)`

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/AppMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.mongodb.stitch.core.internal.common.StitchObjectMapper;
+
+/**
+ * Represents the application metadata for a particular Stitch
+ * application.
+ */
+@SuppressWarnings("unused") // Jackson uses reflection
+public class AppMetadata {
```
adamchel: [nit] in our Java code we generally prefix our classes that represent API models with `Api`, so this could be renamed `ApiAppMetadata`

PR #83, opened 2018-11-28
STITCH-2257 Remote event w/ no version should conflict with pending write w/ no version

adamchel:
Per our discussions in Slack, this fixes the R2L sync routine to not drop events when the remote event version and local pending write version are both empty. This helps prevent us from missing some events that may have happened when both the local and remote document had no version.

To make this work with our tests:
* I fixed an int test to produce a version so that a conflict wouldn't be raised
* I added an int test to test this behavior.
* I fixed two unit tests that now exhibit different behavior. (I still don't completely grasp why I had to make the changes I did, but I think the original tests were wrong and the old behavior was a bug)

### DIFF
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
@@ -759,21 +759,7 @@ public class DataSynchronizer implements NetworkMonitor.StateListener {
     final DocumentVersionInfo lastKnownLocalVersionInfo = DocumentVersionInfo
           .getLocalVersionInfo(docConfig);
 
-    // 1. If both the local document version and the remote change event version are empty, drop
-    //    the event. The absence of a version is effectively a version, and the pending write will
-    //    set a version on the next L2R pass if it’s not a delete.
-    if (!lastKnownLocalVersionInfo.hasVersion() && !currentRemoteVersionInfo.hasVersion()) {
-      logger.info(String.format(
-          Locale.US,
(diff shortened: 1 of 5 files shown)


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1000,7 +986,7 @@ private void syncLocalToRemote() {
 
             // 1. If it does and the version info is different, record that a conflict has occurred.
             //    Difference is determined if either the GUID is different or the version counter is
-            //    greater than the local version counter.
+            //    greater than the local version counter, or if both versions are empty.
```
edaniels: Going to update the spec as well for these changes?

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -1767,8 +1839,18 @@ class SyncIntTestProxy(private val syncTestRunner: SyncIntTestRunner) {
...
+            true -> "delete"
+            false -> event1.fullDocument.toJson()
+        }
+
+        val remoteEventDescription = when (event2.operationType == ChangeEvent.OperationType.DELETE) {
+            true -> "delete"
+            false -> event2.fullDocument.toJson()
+        }
+
+        println("conflict local event: $localEventDescription")
```
edaniels: delete?

adamchel: these were already here, just not null-safe, I'm keeping them because they're useful for debugging

PR #88, opened 2019-01-16
STITCH-1989 Java SDK: Add switch user to auth

jsflax:
This covers:

STITCH-2207
Java SDK: Add ability to log in as another user

STITCH-1987
Java SDK: Add list users to auth

STITCH-1989
Java SDK: Add switch user to auth

Logout/Removal semantics will be the next ticket.

### DIFF
--- a/android/core/src/main/java/com/mongodb/stitch/android/core/auth/StitchAuth.java
+++ b/android/core/src/main/java/com/mongodb/stitch/android/core/auth/StitchAuth.java
@@ -64,7 +64,8 @@ public interface StitchAuth extends Closeable {
   Task<StitchUser> loginWithCredential(final StitchCredential credential);
 
   /**
-   * Logs out the currently logged in user.
+   * Logs out the currently logged in, active user.
+   * Switches to the next logged in user if there is another.
    *
    * @return a {@link Task} completing when logged out.
    */
@@ -78,9 +79,9 @@ public interface StitchAuth extends Closeable {
   boolean isLoggedIn();
 
   /**
-   * Returns the currently logged in user; null if not logged in.
+   * Returns the currently logged in, active user; null if no users are logged in.
    *
-   * @return the currently logged in user; null if not logged in.
+   * @return the currently logged in, active user; null if no users are logged in.
    */
   @Nullable
   StitchUser getUser();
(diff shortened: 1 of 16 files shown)


android/core/src/main/java/com/mongodb/stitch/android/core/auth/StitchAuth.java
```diff
@@ -98,4 +99,20 @@
    * @param listener the listener to remove.
    */
   void removeAuthListener(final StitchAuthListener listener);
+
+  /**
+   * Returns a list of all logged in users.
```
adamchel: [task] add comments to the other methods in this file like `getUser` to clarify that those are operating on the "active" user

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -479,9 +531,9 @@ private StitchUserT processLoginResponse(
       // will keep the currently logged in user logged in if the profile request failed, and in
       // this particular edge case the user is linked, but they are logged in with their older
       // credentials.
```
adamchel: [task] update the comment for this block of code, since it's not just for link requests anymore

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchUserImpl.java
```diff
@@ -69,4 +69,18 @@ public StitchUserProfile getProfile() {
...
+
+  @Override
+  public int hashCode() {
+    return this.id.hashCode();
+  }
+
+  @Override
+  public boolean equals(final Object o) {
+    return o instanceof CoreStitchUser
+        && ((CoreStitchUser) o).getId().equals(getId())
```
adamchel: [style nit] avoid casting many times by doing something like

```java
if (!(o instanceof CoreStitchUser)) {
    return false;
}

const user = (CoreStitchUser) o;
return user.getId().equals(getId()) &&
    ...
```

core/sdk/src/test/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuthUnitTests.java
```diff
@@ -577,6 +580,58 @@ public void testProfileRequestFailureEdgeCases() {
...
+    assertEquals(user, auth.getUser());
+
+    final CoreStitchUser user2 =
+        auth.loginWithCredentialInternal(new UserPasswordCredential("hello ", "friend"));
+
+    assertEquals(user2, auth.getUser());
+    assertNotEquals(user, auth.getUser());
+
+    assertEquals(user, auth.switchUser(user.getId()));
+    assertEquals(user, auth.getUser());
```
adamchel: [task] test to see that the logout case is handled correctly as well. Logout with two logged-in users should set active user to the other user, and logout with one logged-in user should result in no active user, and isLoggedIn to be false.

PR #92, opened 2019-01-30
STITCH-1990 Java SDK: Add reactive user rebinding of services on switch

jsflax:


core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -340,6 +340,11 @@ protected synchronized void logoutInternal(final String userId) throws IllegalAr
         clearUser(authInfo);
         // add logged out user to front of queue
         loggedInUsersAuthInfoList.addFirst(authInfo.loggedOut());
+        try {
+          AuthInfo.writeLoggedInUsersAuthInfoToStorage(loggedInUsersAuthInfoList, storage);
```
jsflax: [bug] driveby

android/core/src/main/java/com/mongodb/stitch/android/core/internal/StitchAppClientImpl.java
```diff
@@ -262,6 +275,41 @@ public boolean isLoggedIn() {
     return getAuth().isLoggedIn();
   }
 
+  @Nullable
+  @Override
+  public String getActiveUserId() {
```
adamchel: [q] what is the advantage of this over just doing `getUser().getId()`?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/CoreRemoteMongoClientImpl.java
```diff
@@ -16,39 +16,55 @@
...
+        appInfo.getNetworkMonitor(),
+        appInfo.getAuthMonitor(),
+        appInfo.getEventDispatcher()
+    );
+  }
+
+  @Override
+  public void onRebindEvent() {
+    // reinitialize the DataSynchronizer entirely.
+    // any auth event will trigger this.
```
adamchel: [minor] the act of adding an auth listener triggers an auth event, meaning this will always be called when starting an app for the first time. I could see that being a problem.

We could refactor the rebinding to only trigger when the user has actually changed, but I could imagine a scenario in the future where an auth event that didn't change the user actually would warrant a service rebinding. 

Maybe it would be best in `DataSynchronizer#reinitialize` to use the `AuthMonitor` to check if the active user id has changed before actually reinitializing. The DataSynchronizer can store the active user whose data is being synchronized.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -296,10 +302,23 @@ public void onNetworkStateChanged() {
     }
   }
 
+  public void reinitialize(final MongoClient localClient) {
+    syncLock.lock();
+    try {
+      this.instanceChangeStreamListener.stop();
+      this.stop();
+      this.localClient = localClient;
+      initialize();
```
adamchel: I think the fact that this is going to be synchronous is a performance issue, but I imagine you will be fixing this in STITCH-2281?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncMongoClientFactory.java
```diff
@@ -30,10 +30,13 @@ public static MongoClient getClient(
     if (dataDir == null) {
       throw new IllegalArgumentException("StitchAppClient not configured with a data directory");
     }
+
+    final String userId = appInfo.getAuthMonitor().isLoggedIn()
+        ? appInfo.getAuthMonitor().getActiveUserId() : "unbound";
     final String instanceKey = String.format(
```
adamchel: [q] do we even need the instance key at all anymore? now that we are storing data for different users in different data directories, I think this is unnecessary. getting rid of it will let users have longer db and coll names as well

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -1686,6 +1687,103 @@ class SyncIntTestProxy(private val syncTestRunner: SyncIntTestRunner) {
         assertNoVersionFieldsInLocalColl(syncTestRunner.syncMethods())
     }
 
+    @Test
+    fun testMultiUserSupport() {
```
adamchel: [task] the comments in this test seem out of date. can you update them to reflect what's actually happening?

.evg.yml
```diff
@@ -357,7 +354,7 @@ tasks:
           SDK_HOME=`pwd`/.android
           $SDK_HOME/platform-tools/adb logcat &
           LOGCAT_PID=$!
-          sleep infinity
+          sleep 86400
```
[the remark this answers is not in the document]
jsflax: `infinity` doesn't work on Darwin systems. 86400 seconds is one day.

android/core/src/main/java/com/mongodb/stitch/android/core/internal/StitchAppClientImpl.java
jsflax: Compartmentalization. No need to expose the developer to fields they do not need.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Yes, this can be fixed in 2281. The fact that our initialization is synchronous currently is already a performance issue, so I'm not worrying about it for this PR.

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
jsflax: Sure.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/CoreRemoteMongoClientImpl.java
jsflax: Good catch. I think it makes more sense to do it here, in the RemoteMongoClient.

jsflax:
Should be of note that removing user data is difficult without revealing the actual auth event.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncMongoClientFactory.java
jsflax: This instance key is actually only used in caching the clients in hashmap.

android/core/src/main/java/com/mongodb/stitch/android/core/internal/StitchAppClientImpl.java
adamchel: I mean the developer might use the other user fields, but I guess now that I think about it, this can be pretty convenient, so it's fine to leave as is

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncMongoClientFactory.java
adamchel: ah okay, cool

adamchel:
Well maybe when we get to that ticket, it might make sense to add a `onUserRemoved()` method to the auth listener. It is an entirely different class of event, and developers may find it useful if they are storing data associated with a user in some other local storage.

android/core/src/main/java/com/mongodb/stitch/android/core/auth/internal/StitchAuthImpl.java
```diff
@@ -189,6 +197,16 @@ public void addAuthListener(final StitchAuthListener listener) {
...
   }
 
+  public void addSynchronousAuthListener(final StitchAuthListener listener) {
+    synchronized (this) {
+      synchronousListeners.add(listener);
+    }
+
+    // Trigger the onUserLoggedIn event in case some event happens and
+    // this caller would miss out on this event other wise.
+    onAuthEvent(listener);
```
dkaminsky: not sure I follow the logic here can you take a minute offline to explain?

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/StitchServiceBinder.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.mongodb.stitch.core.services.internal;
+
+/**
+ * An ifc that allows any service of any type
+ * to bind to it's associated {@link CoreStitchServiceClient}.
+ *
+ * {@link CoreStitchServiceClient#bind(StitchServiceBinder)}
+ */
+public interface StitchServiceBinder {
```
dkaminsky: Could have just added this method with a default no-op to the basic service interface.

PR #94, opened 2019-02-05
STITCH-2281 Java SDK: DataSynchronizer constructor should make blocking initialization asynchronous

jsflax:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -308,8 +307,11 @@ public void reinitialize(final MongoClient localClient) {
       this.instanceChangeStreamListener.stop();
```
adamchel: [nit] is this necessary? It seems like `this.stop()` will call this anyway

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2647,6 +2686,13 @@ public Object call() {
     ).into(new HashSet<>());
   }
 
+  private void assertInitialized() {
```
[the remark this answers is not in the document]
dkaminsky: +1, I don't like assert as a name here

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -303,25 +303,27 @@ public void onNetworkStateChanged() {
...
+      dispatchGroup.unblock();
     }
   }
 
   /**
    * Reloads the synchronization config. This wipes all in-memory synchronization settings.
    */
   public void wipeInMemorySettings() {
-    this.assertInitialized();
+    this.waitUntilInitialized();
```
adamchel: [race condition] another race I can potentially see happening here and with other methods that call this is that: 

1. this `waitUntilInitialized` call is made when `reinitialize` in another thread is on line 309 because a user initiated a user switch
2. `waitUntilInitialized` successfully returns because the initThread is still the old thread, and then this method starts to continue.
3. `reinitialize reaches line 310, which changes the `initThread`, but now those methods that called `waitUntilInitialized` are running even though the data synchronizer is in the middle of initialization.

It's hard to say exactly what the practical implications of this race condition would be, but I could see things happening that would corrupt the sync config and local cache. I think we need to rethink the locking and have some synchronization construct that prevents these `waitUntilInitialized()`-calling functions from running at all while initialization is happening.

And the DispatchGroup doesn't protect you here, because the DispatchGroup gets unblocked before initialization is complete. It's simply to let ongoing operations finish before re-initialization starts I don't have a better solution for you right now but I can try to help think of one.

The only thing that I can think of at the moment that would solve this is always calling `waitUntilInitialized` both before and after enter the dispatch group, and always attempting to enter the dispatch group after the initial `waitUntilInitialized` call, but that seems more complicated than it should be and is probably prone to deadlocks.

Let's spend some more time thinking how to approach this.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -138,7 +138,7 @@ public DataSynchronizer(
     this.listenersLock = new ReentrantLock();
     this.eventDispatcher = eventDispatcher;
     this.instanceKey = instanceKey;
-
+    this.dispatchGroup = new DispatchGroup();
```
adamchel: [naming] I think this should be a bit more descriptive, maybe `ongoingOperationsGroup`?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: This was a simple bug.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -79,6 +79,7 @@ class DataSynchronizerUnitTests {
 
     @After
     fun teardown() {
+        println("teardown called!")
```
adamchel: [nit] remove debug prints

PR #97, opened 2019-02-08
STITCH-2510 Java SDK: Alter AuthListener to better represent User lifecycle

jsflax:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/CoreRemoteMongoClientImpl.java
```diff
@@ -62,21 +64,44 @@ public CoreRemoteMongoClientImpl(final CoreStitchServiceClient service,
...
-      this.dataSynchronizer.reinitialize(
-          SyncMongoClientFactory.getClient(
-              appInfo,
-              service.getName(),
-              clientFactory
-          )
-      );
+          // reinitialize the DataSynchronizer entirely.
+          // any auth event will trigger this.
+          this.dataSynchronizer.reinitialize(
```
adamchel: [q] if there is no active user after the active user has changed, shouldn't we call a different method that just stops the data synchronizer rather than reinitializes it? I see right now that the data synchronizer just ends up being bound to an "unbound" data directory. I guess that's fine but wouldn't it be cleaner if the data synchronizer just didn't run at all when logged out?

server/core/src/main/java/com/mongodb/stitch/server/core/auth/internal/StitchAuthImpl.java
```diff
@@ -90,6 +92,42 @@ public void removeUserWithId(final String userId) {
     removeUserWithIdInternal(userId);
   }
 
+  @Override
+  protected void onListenerInitialized() {
```
adamchel: [q] why do we need to implement these? shouldn't there already be default implementations in the protocol?

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAppClientIntTests.kt
```diff
@@ -32,6 +35,8 @@ import java.util.Arrays
 import java.util.Date
 import java.util.Calendar
 import java.util.concurrent.ExecutionException
+import java.util.concurrent.Semaphore
+import java.util.concurrent.TimeUnit
 
 @RunWith(AndroidJUnit4::class)
 class StitchAppClientIntTests : BaseStitchAndroidIntTest() {
```
adamchel: [task] I think you should probably test more cases auth listener cases. 

check out the JS PR, I added a new file of `StitchAuthListenerIntTests`, and for each method of the listener, I have a test method that simulates the different ways that event can be triggered, and ensures that the event method is called a certain number of times.

it sounds like a lot but it should be pretty simple to write, and it's a lot cleaner than trying to mix in with existing integration tests

android/core/src/main/java/com/mongodb/stitch/android/core/internal/StitchAppClientImpl.java
```diff
@@ -305,11 +307,39 @@ public void onAuthEvent(final StitchAuth auth) {
...
   }
 
+  @Override
+  public void onAuthEvent(StitchAuth auth) {
+  }
+
+  @Override
+  public void onUserLoggedIn(final StitchAuth auth,
+                             final StitchUser loggedInUser) {
+    onRebindEvent(new AuthEvent.UserLoggedIn<>(loggedInUser));
```
tkaye407: Should you be rebinding on all of these?

server/core/src/main/java/com/mongodb/stitch/server/core/auth/internal/StitchAuthImpl.java
jsflax: These are abstract methods on CoreStitchAuth to telegraph to the implementor that it must pass on these methods.

android/core/src/main/java/com/mongodb/stitch/android/core/internal/StitchAppClientImpl.java
jsflax: Yes, it is up to the service to determine what to do with the information given.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/CoreRemoteMongoClientImpl.java
jsflax: It won't run anyway. This won't be an issue until we get to sync cadence. In the interim, this is fine.

jsflax:
@adamchel PTAL

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAuthListenerIntTests.kt
```diff
@@ -0,0 +1,283 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+@RunWith(AndroidJUnit4::class)
+class StitchAuthListenerIntTests : BaseStitchAndroidIntTest() {
```
adamchel: [task] add this class to the server int tests as well

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAuthListenerIntTests.kt
```diff
@@ -0,0 +1,283 @@
...
+            confirmEmailSubject = "email subject",
+            resetPasswordSubject = "password subject")
+        )
+
+        val client = getAppClient(app.first)
+
+        val countDownLatch = CountDownLatch(1)
+
+        client.auth.addAuthListener(object : StitchAuthListener {
+            override fun onAuthEvent(auth: StitchAuth?) {
```
adamchel: [q] can we make `onAuthEvent` default so people don't have to implement it? shouldn't be a breaking change right?

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAuthListenerIntTests.kt
```diff
@@ -0,0 +1,283 @@
...
+
+            override fun onListenerRegistered(auth: StitchAuth?) {
+                assertNotNull(auth)
+                countDownLatch.countDown()
+            }
+        })
+
+        assert(countDownLatch.await(10, TimeUnit.SECONDS))
+    }
+}
```
adamchel: [nit] newline

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/CoreRemoteMongoClientImpl.java
```diff
@@ -77,7 +77,7 @@ private void onAuthEvent(final AuthEvent authEvent) {
           System.err.println("Could not delete database for user id " + userId);
         }
         break;
-      default:
+      case USER_SWITCHED:
```
adamchel: [naming nit] this case should probably be called `ACTIVE_USER_CHANGED`

server/core/src/main/java/com/mongodb/stitch/server/core/auth/internal/StitchAuthImpl.java
```diff
@@ -98,7 +98,7 @@ protected void onListenerInitialized() {
   }
 
   @Override
-  protected void onUserCreated(StitchUser createdUser) {
+  protected void onUserAdded(StitchUser createdUser) {
```
adamchel: [nit q] is this even necessary? we already have a default implementation

core/sdk/src/main/java/com/mongodb/stitch/core/internal/net/EventStreamReader.java
```diff
@@ -94,6 +94,8 @@ protected final Event processEvent() throws IOException {
         }
         doneOnce = true;
         line = "";
+      } catch (IOException e) {
+        return null;
```
adamchel: [q] why not throw an exception? if we're keeping this, update the comment for this method to explain when this may be null, and add a `@Nullable` annotation

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListener.java
```diff
@@ -101,16 +102,25 @@ public void start() {
...
         return;
       }
 
       this.cancel();
+      this.close();
+
       while (runnerThread.isAlive()) {
         runnerThread.interrupt();
         try {
+          System.out.println("killing runner");
```
adamchel: [q] was this for debugging? delete if so. if this would be useful for users, make this an actual log item

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncMongoClientFactory.java
```diff
@@ -33,7 +33,7 @@ public static MongoClient getClient(
       throw new IllegalArgumentException("StitchAppClient not configured with a data directory");
     }
 
-    final String userId = appInfo.getAuthMonitor().isLoggedIn()
+    final String userId = appInfo.getAuthMonitor().tryIsLoggedIn()
```
adamchel: [q] does this need to be a `tryIsLoggedIn`? I'm concerned that this would cause a real user to end up having their data be in an `"unbound"` data directory

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListenerUnitTests.kt
```diff
@@ -40,7 +40,7 @@ class NamespaceChangeStreamListenerUnitTests {
         // assert the stream does not open since we have no document ids
         ctx.isLoggedIn = true
         assertFalse(namespaceChangeStreamListener.openStream())
-        verify(nsConfigMock, times(1)).synchronizedDocumentIds
+        verify(nsConfigMock, times(3)).synchronizedDocumentIds
```
adamchel: [q] what changed to make this happen? this isn't a performance hit right?

server/core/src/main/java/com/mongodb/stitch/server/core/internal/StitchAppClientImpl.java
```diff
@@ -181,8 +182,17 @@ public void callFunction(
   }
 
   @Override
-  public boolean isLoggedIn() {
-    return getAuth().isLoggedIn();
+  public boolean isLoggedIn() throws InterruptedException {
```
adamchel: [q] If a user ever calls `isLoggedIn()` method on a `StitchAppClient`, is there a case where that can throw? should we mark the method as `throws` in the interface and explain when it could happen?

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -267,9 +297,15 @@ private synchronized Response doAuthenticatedRequest(
...
                                                final Decoder<T> decoder) {
-    if (!isLoggedIn()) {
-      throw new StitchClientException(StitchClientErrorCode.MUST_AUTHENTICATE_FIRST);
+    try {
+      if (!isLoggedInInterruptibly()) {
+        throw new StitchClientException(StitchClientErrorCode.MUST_AUTHENTICATE_FIRST);
+      }
+    } catch (InterruptedException e) {
+      System.err.println("Stream authentication was interrupted.");
+      return null;
```
adamchel: [q] are we sure we want to return this? why not just throw the exception? This seems like a null pointer exception waiting to happen somewhere upstream. What happens if the user calls `watch` and the stream got interrupted? I think it would be better for an error to happen rather than getting a null result.

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -267,9 +297,15 @@ private synchronized Response doAuthenticatedRequest(
...
   public <T> Stream<T> openAuthenticatedStream(final StitchAuthRequest stitchReq,
                                                final Decoder<T> decoder) {
-    if (!isLoggedIn()) {
-      throw new StitchClientException(StitchClientErrorCode.MUST_AUTHENTICATE_FIRST);
+    try {
+      if (!isLoggedInInterruptibly()) {
+        throw new StitchClientException(StitchClientErrorCode.MUST_AUTHENTICATE_FIRST);
+      }
+    } catch (InterruptedException e) {
+      System.err.println("Stream authentication was interrupted.");
```
adamchel: make this a log statement or get rid of the print

core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
jsflax: At the moment, that would have to happen on the users accord. We should probably throw an exception.

server/core/src/main/java/com/mongodb/stitch/server/core/internal/StitchAppClientImpl.java
jsflax: No, all of this is internal.

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListenerUnitTests.kt
jsflax: Hmm... it is a slight performance hit. I moved `getSynchronizedDocumentIds` to the top of `openStream` to pre-read them before taking the lock. But that may not be necessary anymore.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncMongoClientFactory.java
jsflax: In the current state of our codebase, it would not be possible for this to happen. However, maybe I'll make a note here for "in the future".

server/core/src/main/java/com/mongodb/stitch/server/core/internal/StitchAppClientImpl.java
adamchel: ah you're right, there's no isLoggedIn on the app client itself

PR #98, opened 2019-02-11
STITCH-1986 - Add watch to remote mongo collection

dkaminsky:
This review consists of the work necessary to expose the remote collection's `watch` function to end users, including conversion of the `BsonDocument `representation of the full document as provided by the remote change stream into the user's local object representation. This is exposed through two new methods on the remote collection, one which accepts a number of `BsonValue` identifiers and one which accepts the non-BSON `ObjectId` type (for convenience).


android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoCollection.java
```diff
@@ -241,6 +246,22 @@
...
+   * @return the stream of change events.
+   */
+  Task<Stream<ChangeEvent<DocumentT>>> watch(final ObjectId... ids);
+
+  /**
+   * Watches specified IDs in a collection.
+   * @param ids the ids to watch.
+   * @return the stream of change events.
+   */
+  Task<Stream<ChangeEvent<DocumentT>>> watch(final BsonValue... ids);
```
jsflax: [major] This pattern will not play nicely on Android.

A traditional listener pattern here would be more idiomatic. Core `Stream`'s `nextEvent` is also blocking, which is a no-go on Android.

What we're doing on iOS (and similarly JS) is to return a "session" that can invalidate the stream if need be, and that session acts as a shim between the delegate/listener and the stream itself.

Something like:
```java
WatchSession watch(final WatchChangeEventListener listener, final BsonValue... ids);

interface WatchChangeEventListener extends ChangeEventListener {
    void onStreamOpen();
    void onStreamClosed();
}

interface WatchSession {
    void invalidateStreamAndClose();
}

class WatchSessionImpl<DocumentT> implements WatchSession {
   WatchSessionImpl(Stream<DocumentT> stream, WatchChangeEventListener listener) {
       ....
   }
} 
```

The current `Stream` type would also need to be more reactive like Swift–  this is a flaw in the original design. The stream should have callbacks for `onOpen`, `onClosed`, `onEvent`. These should be bubbled up to the session, which then dispatches it to the listeners.

This is going to be required for the Sync cadence ticket as well. Streaming should be entirely asynchronous.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -88,9 +87,10 @@
  * and a remote MongoDB (via Stitch). It also expose CRUD operations to interact with synchronized
  * documents.
  */
+//TODO: API24 replace deprecated forEach calls with forEach(Consumer)
```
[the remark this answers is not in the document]
dkaminsky: I'll remove, that line wasn't supposed to be committed.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/OperationType.java
```diff
@@ -0,0 +1,66 @@
...
+        return "delete";
+      case REPLACE:
+        return "replace";
+      case UPDATE:
+        return "update";
+      default:
+        return "unknown";
+    }
+  }
+}
```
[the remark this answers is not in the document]
dkaminsky: done

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoCollection.java
dkaminsky: done -- callbacks aren't as idiomatic for Java as they are for Swift, so I've reworked the stream to return new `Task`s for each event instead of synchronously returning events for now -- a callback mechanism can be built on top of this at the app level for now, but I have a feeling this all may need to be reworked anyway for reactive syncing

.evg.yml
```diff
@@ -327,7 +327,7 @@ tasks:
             cd stitch-java-sdk
             echo "running android tests"
             echo "test.stitch.baseURL=http://10.0.2.2:9090" >> local.properties
-            ./gradlew connectedAndroidTest --info --continue --warning-mode=all --stacktrace  < /dev/null
+            ./gradlew connectedAndroidTest jacocoTestReport --info --continue --warning-mode=all --stacktrace  < /dev/null
```
adamchel: [nit] this did not seem to work. let's get rid of this for now and file a ticket to fix android coverage

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
```diff
@@ -64,7 +64,7 @@ void configure(@NonNull final ConflictHandler<DocumentT> conflictHandler,
    * Requests that the given document _ids be synchronized.
    * @param ids the document _ids to synchronize.
    */
-  void syncMany(final BsonValue... ids);
+  void   syncMany(final BsonValue... ids);
```
adamchel: [nit] too much whitespace

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+
+import org.bson.BsonValue;
+
+/**
+ * User-level abstraction for a stream returning {@link ChangeEvent}s from the server.
+ *
+ * @param <UserT> The type returned to users of this API when the next event is requested.
```
adamchel: [naming] I think EventT might be a better name for this. UserT makes it sound like it's a type representing a user.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ExceptionListener.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.stitch.core.services.mongodb.remote;
+
+import org.bson.BsonValue;
+
+/**
+ * ExceptionListener receives non-network related errors that occur.
```
adamchel: [q] are we 100% sure this is only for non-network errors? I could see a corrupted event over the network causing an error here.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/ResultDecoders.java
```diff
@@ -171,7 +171,14 @@ public boolean equals(final Object obj) {
 
       // caveat: if someone writes a stateful codec then this logic won't hold up, but we
       // can't use .equals without opening a can of worms
+      System.err.println("FOOOO(this): " + this.codec.getClass().getCanonicalName());
```
adamchel: [nit] remove debug prints

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ExceptionListener.java
adamchel: ah I see you just copied this from the previous ErrorListener. let's leave for now but we can discuss offline

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
```diff
@@ -1868,14 +1870,16 @@ class DataSynchronizerUnitTests {
 
         ctx.dataSynchronizer.syncDocumentsFromRemote(ctx.namespace, ctx.testDocumentId)
 
-        ctx.waitForDataSynchronizerStreams();
+        ctx.waitForDataSynchronizerStreams()
 
         ctx.doSyncPass()
 
         val mockEmptyFindResult = mock(CoreRemoteFindIterableImpl::class.java)
+        @Suppress("UNCHECKED_CAST")
```
adamchel: [q] why did we have to add these?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1703,14 +1702,17 @@ public void syncDocumentsFromRemote(
       final BsonValue... documentIds
   ) {
     this.waitUntilInitialized();
+    boolean added = false;
 
     try {
       ongoingOperationsGroup.enter();
       for (final BsonValue documentId : documentIds) {
-        syncConfig.addSynchronizedDocument(namespace, documentId);
+        added = added || syncConfig.addSynchronizedDocument(namespace, documentId);
```
adamchel: [bug] I think short-circuit boolean logic means that if `added` is true, then the document won't be added to the syncConfig, meaning that if you try to sync multiple ids, it will ignore all the ids after the first added one.

I think this should be 

```java
final boolean docAdded = syncConfig.addSynchronizedDocument(namespace, documentId);
added = added || docAdded;
```

you could also probably rename `added` to `anyDocsAdded` or `someDocsAdded`

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1730,20 +1732,26 @@ public void desyncDocumentsFromRemote(
     this.waitUntilInitialized();
     final Lock lock = this.syncConfig.getNamespaceConfig(namespace).getLock().writeLock();
     lock.lock();
+    boolean removed = false;
     try {
       ongoingOperationsGroup.enter();
+
       for (final BsonValue documentId : documentIds) {
-        syncConfig.removeSynchronizedDocument(namespace, documentId);
+        removed = removed || syncConfig.removeSynchronizedDocument(namespace, documentId);
```
adamchel: [bug] same issue as above

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: Oh good catch I need non-short circuiting behavior

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerUnitTests.kt
dkaminsky: silencing Kotlin warnings

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
dkaminsky: done

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/Sync.java
dkaminsky: done

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ExceptionListener.java
dkaminsky: discussed offline

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/internal/ResultDecoders.java
dkaminsky: done

.evg.yml
dkaminsky: looking into this a little more before removing

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: ditto

PR #100, opened 2019-02-15
Small fixes for multi-user auth logic

adamchel:


core/sdk/src/main/java/com/mongodb/stitch/core/auth/internal/CoreStitchAuth.java
```diff
@@ -355,10 +355,6 @@ public StitchUserT switchToUserWithId(final String userId)
   protected StitchUserT loginWithCredentialInternal(final StitchCredential credential) {
     authLock.writeLock().lock();
     try {
```
tkaye407: lol, good catch

PR #107, opened 2019-03-01
STITCH-2573 - Close streams on user switch

dkaminsky:
Whenever we create a stream we need to close that stream when a user switches, whether the stream was created manually by the user or not. This change keeps a weak reference set of `allocatedStreams` to be canceled/closed when active user changes and implements the change. Tests have been added to ensure that a rebind event causes the core service client to do this, omitting any already-closed streams.

Also added an addiitonal test on core service client to make sure that rebind event is dispatched to all binders.


gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
tkaye407: How come you added this?

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
```diff
@@ -195,14 +202,44 @@ public CoreStitchServiceClient withCodecRegistry(final CodecRegistry codecRegist
 
   @Override
   public void bind(final StitchServiceBinder binder) {
-    this.serviceBinders.add(new WeakReference<>(binder));
+    this.serviceBinders.put(new WeakReference<>(binder), Boolean.TRUE);
```
adamchel: [nit] can you not just pass `true` here?

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
```diff
@@ -176,9 +181,11 @@ public void callFunction(
   public <T> Stream<T> streamFunction(final String name,
                                       final List<?> args,
                                       final Decoder<T> decoder) throws InterruptedException {
-    return requestClient.openAuthenticatedStream(
+    final Stream<T> newStream = requestClient.openAuthenticatedStream(
         getStreamServiceFunctionRequest(name, args), decoder
     );
+    this.allocatedStreams.put(new WeakReference<>(newStream), Boolean.TRUE);
```
adamchel: [nit] ditto

gradlew
dkaminsky: generated by gradle wrapper task

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
```diff
@@ -195,14 +202,44 @@ public CoreStitchServiceClient withCodecRegistry(final CodecRegistry codecRegist
...
+    switch (rebindEvent.getType()) {
+      case AUTH_EVENT:
+        isActiveUserChange = ((AuthEvent) rebindEvent).getAuthEventType() ==
+            AuthEvent.Type.ACTIVE_USER_CHANGED;
+        break;
+      default:
+        break;
+    }
+    if (isActiveUserChange) {
+      final Iterator<WeakReference<Stream<?>>> streamIterator =
```
adamchel: [style nit] just for the sake of not doing too many things in one method, I would put the stuff in this conditional in a method called `closeAllocatedStreams()` and just call it here

core/sdk/src/test/java/com/mongodb/stitch/core/internal/net/StreamTestUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    return doCreateStream(decoder, true, streamContent);
+  }
+
+  public static <T> Stream<T> createClosedStream(final Decoder<T> decoder,
+                                                 final String... streamContent) throws IOException {
+    return doCreateStream(decoder, false, streamContent);
+  }
+
+  private static <T> Stream<T> doCreateStream(final Decoder<T> decoder,
+                                            final boolean open,
```
adamchel: [nit] vertical parameter alignment

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
dkaminsky: You can't satisfy generics with primitive types in Java, so it would be auto-wrapped to the equivalent of this anyway, so it's easier to use the constant.

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
dkaminsky: ditto

core/sdk/src/test/java/com/mongodb/stitch/core/internal/net/StreamTestUtils.java
dkaminsky: got it

core/sdk/src/main/java/com/mongodb/stitch/core/services/internal/CoreStitchServiceClientImpl.java
dkaminsky: fair enough

PR #111, opened 2019-03-11
STITCH-2596 - PMD configuration and driveby of some 20-30 violations

dkaminsky:


core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
```diff
@@ -17,57 +17,78 @@
...
+    }
     if (dbPath == null || dbPath.isEmpty()) {
       throw new IllegalArgumentException("dbPath must be non-empty");
     }
-    if (instances.containsKey(key)) {
-      return instances.get(key);
-    }
 
-    for (final MongoClient value : instances.values()) {
-      value.close();
```
tkaye407: Seems like strange repetition below with the same if statements. I feel like you should be able to take out the first if statement, but I am not as familiar with concurrent Java programming

server/core/src/main/java/com/mongodb/stitch/server/core/Stitch.java
```diff
@@ -58,14 +63,16 @@ public static synchronized StitchAppClient getDefaultAppClient() {
...
-    if (!appClients.containsKey(clientAppId)) {
-      throw new IllegalStateException(
-          String.format("client for app '%s' has not yet been initialized", clientAppId));
+    synchronized (Stitch.class) {
+      if (!appClients.containsKey(clientAppId)) {
+        throw new IllegalStateException(
+            String.format("client for app '%s' has not yet been initialized", clientAppId));
+      }
+      return appClients.get(clientAppId);
     }
```
tkaye407: Same comment as before regarding the necessity of this synchronized block

core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
```diff
@@ -17,57 +17,78 @@
 package com.mongodb.stitch.core.services.mongodb.local.internal;
 
 import com.mongodb.client.MongoClient;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import org.bson.codecs.configuration.CodecRegistry;
 
 public abstract class EmbeddedMongoClientFactory {
-
```
tkaye407: Im not super familiar with how this code used to work, but it seems strange that we are no longer caring about storing other MongoClient instances. Are we just saying that you should only have one at a time?

android/core/src/main/java/com/mongodb/stitch/android/core/Stitch.java
```diff
@@ -40,22 +42,28 @@
   private static final String DEFAULT_BASE_URL = "https://stitch.mongodb.com";
   private static final Long DEFAULT_DEFAULT_REQUEST_TIMEOUT = 60000L;
   private static final String TAG = Stitch.class.getSimpleName();
-  private static final Map<String, StitchAppClientImpl> appClients = new HashMap<>();
-  private static boolean initialized;
+  private static final ConcurrentMap<String, StitchAppClientImpl> appClients =
+      new ConcurrentHashMap<>();
+  private static final AtomicBoolean initialized = new AtomicBoolean(false);
+  private static volatile String defaultClientAppId;
```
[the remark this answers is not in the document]
dkaminsky: the concurrent hash map uses a bucket-level locking scheme to efficiently provide locking concurrency so multiple threads can use the data structure at the same time with minimal blocking

atomic boolean uses an atomic "compare and swap" operation at the native level to provide  thread safety and visibility guarantees around the value, so you don't need to use separate `synchronize`-style locks around its usage

volatile just guarantees visibility between threads, e.g. if I change it in thread A then thread B sees it has changed on the next instruction - it's not actually providing thread safety but prevents multiple threads from seeing different values for the same variable

android/core/src/main/java/com/mongodb/stitch/android/core/Stitch.java
```diff
@@ -103,15 +111,18 @@ public static synchronized StitchAppClient getDefaultAppClient() {
...
-    if (!appClients.containsKey(clientAppId)) {
-      throw new IllegalStateException(
-          String.format("client for app '%s' has not yet been initialized", clientAppId));
+
+    synchronized (Stitch.class) {
+      if (!appClients.containsKey(clientAppId)) {
+        throw new IllegalStateException(
+            String.format("client for app '%s' has not yet been initialized", clientAppId));
+      }
+      return appClients.get(clientAppId);
```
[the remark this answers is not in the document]
dkaminsky: because the above line is checking ```if (!appClients.containsKey(clientAppId))```, if another thread performs a destructive operation in between that check and the return, a user could receive a null result, which is not our current contract for that method

core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
dkaminsky: Example:
```1:   if (someCondition) {
2:        synchronized(someObject) {
3:            doSomething();
4:        }
5:    }

Suppose that two threads hit line 1 at the same time, they will both evaluate the same result for `someCondition`. Now supposing `someCondition` is true, they will both attempt to lock on `someObject`.  One will succeed and proceed to perform `doSomething()` while the other blocks and waits for the lock on `someObject` to become available. When the winning thread exits the `synchronized` block the losing thread will then enter it and also perform `doSomething()`

If `doSomething` is idempotent then it doesn't matter, but if `doSomething` has side effects then this is broken. The double-check pattern, though imperfect, prevents the losing thread from duplicating the inner operation.

core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
dkaminsky: The previous code, in practice, only stored a maximum of one key. This is because in current state, we can only have one active client at any given time and the rest must be aggressively closed.

android/core/src/main/java/com/mongodb/stitch/android/core/Stitch.java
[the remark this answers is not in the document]
tkaye407: Would my suggestion above behave any differently though?

core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
tkaye407: Got it

core/services/mongodb-local/src/main/java/com/mongodb/stitch/core/services/mongodb/local/internal/EmbeddedMongoClientFactory.java
tkaye407: I understand why you need the inner condition, but I was more confused about why you would need the outer one. I guess in the case that someCondition is false you avoid entering the synchronized block at all?

PR #115, opened 2019-03-13
STITCH-2584 Java SDK: Use bulk operations wherever possible on local database for sync

jsflax:


android/build.gradle
```diff
@@ -32,7 +32,8 @@ subprojects {
                     if (file.exists()) {
                         properties.load(project.rootProject.file('local.properties').newDataInputStream())
                     }
-                    testInstrumentationRunnerArgument "test.stitch.baseURL", properties.getProperty("test.stitch.baseURL", "http://10.0.2.2:9090")
+                    testInstrumentationRunnerArgument "test.stitch.baseURL",
+                            properties.getProperty("test.stitch.baseURL", "http://10.1.14.181:9090")
```
edaniels: Undo

android/examples/stress-tests/build.gradle
```diff
@@ -25,4 +35,9 @@ dependencies {
     implementation "com.android.support:appcompat-v7:${support_library_version}"
     implementation "com.android.support:recyclerview-v7:${support_library_version}"
     implementation "com.android.support.constraint:constraint-layout:1.1.0"
+//    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.1.1'
```
edaniels: uncomment?

android/services/mongodb-remote/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,4 @@
-<manifest package="com.mongodb.stitch.android.services.mongodb.remote">
+<manifest package="com.mongodb.stitch.android.services.mongodb.remote"
+          xmlns:android="http://schemas.android.com/apk/res/android">
+    <application android:usesCleartextTraffic="true" />
```
edaniels: I think you can undo these changes

android/testutils/src/main/java/com/mongodb/stitch/android/testutils/BaseStitchAndroidIntTest.kt
```diff
@@ -36,7 +36,7 @@ open class BaseStitchAndroidIntTest : BaseStitchIntTest() {
     }
 
     override fun getStitchBaseURL(): String {
-        return InstrumentationRegistry.getArguments().getString("test.stitch.baseURL", "http://10.0.2.2:9090")
+        return "http://10.0.2.2:9090"
```
edaniels: Should this be reverted?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceSynchronizationConfig.java
```diff
@@ -227,6 +236,32 @@ Codec getDocumentCodec() {
     return documentCodec;
   }
 
+  boolean addSynchronizedDocuments(
+      final BsonValue... documentIds
+  ) {
+    Map<BsonValue, CoreDocumentSynchronizationConfig> configs = new HashMap<>();
+    for (final BsonValue documentId : documentIds) {
+      if (getSynchronizedDocument(documentId) == null) {
```
edaniels: [q] Should we add a `getSynchronizedDocuments` to speed this up?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceSynchronizationConfig.java
```diff
@@ -250,6 +285,32 @@ boolean addSynchronizedDocument(
     }
   }
 
+  @Nullable
+  DeleteManyModel<CoreDocumentSynchronizationConfig> removeSynchronizedDocuments(
```
edaniels: [nit] Maybe could call this `prepareRemoveSynchronizedDocuments` so as to indicate this won't remove them at this call site

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/InstanceSynchronizationConfig.java
```diff
@@ -162,6 +164,14 @@ public CoreDocumentSynchronizationConfig addAndGetSynchronizedDocument(
     return nsConfig.getSynchronizedDocument(documentId);
   }
 
+  @Nullable
+  DeleteManyModel<CoreDocumentSynchronizationConfig> removeSynchronizedDocuments(
```
edaniels: Similar comment about prepare

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -91,6 +97,75 @@
  * documents.
  */
 public class DataSynchronizer implements NetworkMonitor.StateListener {
+  class BatchOps {
```
edaniels: [nit] May be better to move this into its own class in a different file instead of this being an inner non-static class. Would just need to provide more args for its methods though to compensate not having direct access. Might be cleaner though.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -91,6 +97,75 @@
...
+    }
+
+    BatchOps() {
+      this.bulkWriteModels = new ArrayList<>();
+      this.configs = new ArrayList<>();
+      this.ids = new HashSet<>();
+    }
+
+    void merge(@Nullable final BatchOps batchOps) {
+      if (batchOps != null) {
```
edaniels: [nit] Can make this if null early return to reduce indentation

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2523,6 +2613,7 @@ public BsonValue apply(@NonNull final BsonDocument bsonDocument) {
           undoCollection.deleteOne(getDocumentIdFilter(documentId));
           eventsToEmit.add(event);
         }
+        triggerListeningToNamespace(namespace);
```
edaniels: [q] Do we need to trigger here as a part of this change?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -228,6 +267,16 @@ void setSomePendingWrites(
     }
   }
 
+  void setPendingWritesCompleteNoDB(final BsonDocument atVersion) {
```
edaniels: [q] Is the point of this to make sync not skip over these during batch processing?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -127,12 +143,12 @@ public boolean isStale() {
   public void setStale(final boolean stale) {
     docLock.writeLock().lock();
     try {
-      docsColl.updateOne(
-          getDocFilter(namespace, documentId),
-          new BsonDocument("$set",
-              new BsonDocument(
-                  CoreDocumentSynchronizationConfig.ConfigCodec.Fields.IS_STALE,
-                  new BsonBoolean(stale))));
+//      docsColl.updateOne(
```
edaniels: Should this be uncommented or method removed?

android/examples/stress-tests/build.gradle
jsflax: Can just delete, we don't need core.

android/testutils/src/main/java/com/mongodb/stitch/android/testutils/BaseStitchAndroidIntTest.kt
jsflax: Yes

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceSynchronizationConfig.java
jsflax: I don't think we need to, it's just reading from a hashmap.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
jsflax: Removed

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
jsflax: Yes, `desync` used to trigger every time it was called. It no longer does that, so we need to do it manually after desyncing has been done. There is a cleaner way to do this by tracking what operations have been done and if the streams need to be restarted.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
jsflax: The point of this is to not replace the doc config everytime we set pending writes– that's very slow. So we change it in memory here and do a bulk write after the pass on the namespace is done.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceSynchronizationConfig.java
edaniels: :+1:

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
edaniels: cool

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -127,12 +140,6 @@ public boolean isStale() {
   public void setStale(final boolean stale) {
     docLock.writeLock().lock();
     try {
-      docsColl.updateOne(
-          getDocFilter(namespace, documentId),
-          new BsonDocument("$set",
-              new BsonDocument(
-                  CoreDocumentSynchronizationConfig.ConfigCodec.Fields.IS_STALE,
-                  new BsonBoolean(stale))));
       isStale = stale;
```
dkaminsky: [q] approve of caching locally but looks like this never gets written to the docConfig as a result - are we ok with this?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -228,15 +235,34 @@ void setSomePendingWrites(
...
 
+  /**
+   * Sets that there are some pending writes that occurred at a time for an associated
+   * locally emitted change event. This variant updates the last version set.
+   *
+   * @param atTime      the time at which the write occurred.
+   * @param atVersion   the version for which the write occurred.
+   * @param changeEvent the description of the write/change.
+   */
+  void setSomePendingWrites(
```
dkaminsky: [q] why add this? does not appear this is called anywhere? did I miss it?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
jsflax: It does get written when we do the bulk writes.

In the event that the app crashes before it can be written, the document is marked stale anyway on app restart.

PR #116, opened 2019-03-13
DOCSP-4586 Add feedback widget to hosted docs

cbush:
https://jira.mongodb.org/browse/DOCSP-4586

### Summary

This adds the feedback ~and intercom~ widgets to the hosted docs
(more specifically: docs generated with the 'analytics' enabled).

Like the analytics snippet, these widget snippets and some supporting
css, js and HTML pieces are hacked into the header html that javadoc's
standard doclet allows us to include on each page.

### Apologia, or: Oh, Javadoc
Javadoc's standard doclet supports multiple additional css files
using the --add-stylesheet flag, but the options object in the build.gradle
script only allows one additional stylesheet. Besides, there doesn't
seem to be a way to include non-css supporting files (i.e. additional
js files) in each page. Therefore, I resorted to importing some of the
supporting stylesheets and js files from the docs site for the JS sdk.
(Since this header is only included for live docs anyway, this should
not impact local/offline users of the docs at all.)

Bringing in the complete style assets as seen on the JS sdk docs
is too heavy at this time and doesn't play nicely with the standard
doclet styling. So a few required css properties are inlined in HTML
and the font-awesome part of the overall style (just for the
thumbs-up, thumbs-down and chat bubble icons) is extracted and built
as a standalone file. Instructions on how to do that are in the
contrib/assets/css/README.md file -- I deemed bringing in the entire
theme build just for a couple icons to be overkill. Hopefully, this
is only a temporary solution as we have plans to overhaul the javadoc
output anyway.

Finally, the fonts need to be copied into the javadoc directory
but I can't get this to work as a gradle task. The javadoc step
always happens after the task (even with, e.g.
`copyDocFonts.mustRunAfter doAllJavadocs`). So I just put it in
the generate_docs.sh script instead.

### Screenshots
image

image

### Staged

https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docs-feedback/sdk/java/index.html


config/header.html
```diff
@@ -7,3 +8,24 @@
   });
   }}();
 </script>
+<style>
+.deluge {
+  background-color: #13aa52;
+}
+.deluge .switch.good.selected, .deluge .switch.good.selected:hover {
+  color: #13aa52;
+}
```
cbush: These are inlined, as reporting from relative paths is nigh impossible and I can only add one additional css file via javadoc.

PR #124, opened 2019-03-29
STITCH-2652 mobile sync performance test definitions for L2R-only scenarios

adamchel:
This adds the performance test definitions for the L2R-only scenarios.

In the scenarios, you'll see that I used `beforeEach` to set up the test completely so that setup is not measured as part of the perf test, and that I used `afterEach` to verify that the test ran correctly. The test definition itself should only be running the actual scenario to be tested.

This took a while just to get everything hooked up correctly. The remainder of the definitions (R2L-only, and R2L/L2R mixed) should be a bit more straightforward to write.

**_Drive-bys:_**
* Fix perf test harness to properly work in a Mobile Sync environment
    * required hooking up the correct `testNetworkMonitor` to the `outputStitchClient`
    * required not tearing down our embedded mongo client after each test
* In perf test harness, handle errors in test execution by reporting failed runs rather than dropping the entire result
* Fix linter for perf test harness
* Deleted the performance testing files that were copied over from the remote mongodb service package in Jason's PR


android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncPerformanceIntTestsHarness.kt
```diff
@@ -170,14 +175,20 @@ class SyncPerformanceIntTestsHarness : BaseStitchAndroidIntTest() {
...
                         Tasks.await(outputColl.updateOne(filterDocument, updateDocument))
                     }
                 }
             }
         } catch (e: Exception) {
-            Tasks.await(outputColl.deleteOne(Document("_id", resultId)))
-            throw e
+            Tasks.await(outputColl.updateOne(
+                    Document("_id", resultId),
+                    Document("\$set", Document()
```
adamchel: @tkaye407 I can modify this whenever your approach for iOS is set in stone

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncPerformanceTestContext.kt
```diff
@@ -112,15 +113,18 @@ class SyncPerformanceTestContext(private val harness: SyncPerformanceIntTestsHar
             Tasks.await(testColl.deleteMany(Document()))
         }
 
-        if (::testMongoClient.isInitialized) {
+        // TODO: The test should be set up in such a way that each setup sets up an embedded mongo
```
adamchel: I can create a ticket for this, but I think this should really be a consequence of just doing general QoL improvements to the performance test harness

android/build.gradle
```diff
@@ -10,15 +10,15 @@ buildscript {
...
-    if (env['com.android.studio.gradle.project.path'] == null) {
-        project.tasks.configureEach {
-            it.onlyIf {
-                project.hasProperty("perf") ?
-                        project.name.contains("perftest") :
-                        !project.name.contains("perftest")
-            }
-        }
-    }
+//    if (env['com.android.studio.gradle.project.path'] == null) {
```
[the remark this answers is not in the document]
tkaye407: So are we leaving this in here commented out?

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncPerformanceIntTestsHarness.kt
tkaye407: Ill go in and make some changes to the android harness once Im done with iOS

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncL2ROnlyPerformanceTests.kt
```diff
@@ -0,0 +1,322 @@
...
+import com.mongodb.stitch.core.services.mongodb.remote.sync.DefaultSyncConflictResolvers
+import org.bson.Document
+import org.bson.types.ObjectId
+import org.junit.Test
+import kotlin.random.Random
+
+import com.mongodb.stitch.core.services.mongodb.remote.ExceptionListener
+import org.junit.After
+import org.junit.Before
+
```
tkaye407: Do we want tests to be in different files? In that case the runId's will be different for every test which is not what we wanted. ** Adam and I spoke offline about perhaps keeping the logic of this class and then having a devoted test file that runs the tests and calls the helper functions defined in here

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListener.java
```diff
@@ -254,7 +254,7 @@ void storeNextEvent() {
           return;
         }
 
-        logger.info(String.format(Locale.US,
+        logger.debug(String.format(Locale.US,
             "NamespaceChangeStreamListener::stream ns=%s event found: op=%s id=%s",
```
tkaye407: Seems like a strange drive-by so just making sure you meant to make this change

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncPerformanceIntTestsHarness.kt
tkaye407: Yeah currently this is not the best behaviour because a failure will stop all tests but I can fix it later=

android/build.gradle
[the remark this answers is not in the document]
adamchel: Will let @jsflax respond, I'm not quite sure what to do with it

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncL2ROnlyPerformanceTests.kt
adamchel: I'm happy to have a single class, and abstract the L2R stuff in a proxy file that we'll base the other scenario files on. I'll do that when addressing your comments.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/NamespaceChangeStreamListener.java
adamchel: Yeah I did this because the log was extremely noisy when getting thousands of events.

android/build.gradle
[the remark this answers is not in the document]
jsflax: Yep, my fault. You can remove this block for now and I'll update the README later on how to run perf tests.

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncL2ROnlyPerformanceTests.kt
jsflax: I do not think we should have a single class. There are simple ways to share a runId if need be. We don't need another monolith, not to mention the OOM issues we've been facing.

adamchel: by the time I saw this, it had already been done... but I did it in such a way that the definitions are in separate files, so I wouldn't quite call it a monolith. If we run into the OOM issue again, we can fix this up by having each test definition class handle its own test harness

android/services/mongodb-remote-perftests/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/SyncL2ROnlyPerformanceTestDefinitions.kt
```diff
@@ -0,0 +1,298 @@
...
+            val params = TestParams(
+                    runId = runId,
+                    testName = testName,
+                    dataProbeGranularityMs = 400L,
+                    docSizes = docSizes,
+                    numDocs = numDocs,
+                    numIters = 3,
+                    numOutliersEachSide = 0,
+                    outputToStitch = true,
+                    stitchHostName = "https://stitch.mongodb.com"
```
jsflax: [q] shouldn't this be configurable?

adamchel: sure. I'll make this another local property like the Stitch API key. honestly all of these should be fed into local.properties, but we can iterate as we add more test definitions.

PR #127, opened 2019-04-03
DOCSP-4911 Add code example snippets as integration tests

cbush:
These can be copied over to the contrib/docs-examples directory manually for now
upon change.

This adds a new module under the examples module called "snippets" which is
just a bunch of integration tests containing the code examples.

### DIFF
--- a/android/examples/snippets/src/main/AndroidManifest.xml
+++ b/android/examples/snippets/src/main/AndroidManifest.xml
@@ -1,0 +1,1 @@
+<manifest package="com.mongodb.stitch.android.examples.snippets" />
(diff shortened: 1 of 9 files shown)


contrib/README.md
```diff
@@ -49,3 +49,14 @@ The general publishing flow can be followed using `major` as the bump type in `b
     * You must run at least one ```mongod``` instance with replica sets initiated or a ```mongos``` instance with same locally on port 26000
     * You must run the Stitch server locally using the Android-specific configuration:
         ```--configFile ./etc/configs/test_config_sdk_base.json --configFile ./etc/configs/test_config_sdk_android.json```
+
+Hint:
```
adamchel: [nit] Hint? can you be more clear about what this is for?

android/examples/snippets/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1 @@
+<manifest package="com.mongodb.stitch.android.services.mongodb.remote" />
```
adamchel: [q] should this be be `"com.mongodb.stitch.android.examples.snippets"`?

cbush: Yep, good catch!

PR #128, opened 2019-04-03
STITCH-2664 Java SDK: Fix intermittent OOM on Evergreen

jsflax:
Updates and fixes to evergreen

- Builds only take ~10-11 minutes
- There is now an individual task for each module and lint task
- DS Unit Tests have been split up to skirt the OOM issue
- Tasks will be significantly easier to debug
- I’ve had to temporarily turn off coveralls


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncWriteModelContainer.java
```diff
@@ -69,7 +69,7 @@ void wrapForRecovery(final MongoCollection<BsonDocument> localCollection,
                        final MongoCollection<BsonDocument> undoCollection,
                        final Runnable callable) {
     final List<BsonDocument> oldDocs = localCollection.find(
-        new Document("_id", new Document("$in", ids))
+        new BsonDocument("_id", new BsonDocument("$in", new BsonArray(new ArrayList<>(ids))))
```
adamchel: [q] just curious, what was this drive-by for?

.evg.yml
```diff
@@ -324,174 +405,339 @@ tasks:
...
+    tasks:
+      - findbugs
+      - lint
+      - ktlint
+      - checkstyle
+      - pmd
 buildvariants:
-- name: macos-1014-stitch
-  display_name: macos-1014-stitch
+- name: macos-1014
```
adamchel: [q] why not macos-1014-stitch? If the server team ever gets hit with a bunch of builds all at once, this could delay our task significantly

.evg.yml
```diff
@@ -324,174 +405,339 @@ tasks:
...
+          script: |
+            set -e
+            sh stitch-java-sdk/etc/run_core_test.sh :android:android-services:stitch-android-services-twilio:connectedAndroidTest
+    post:
+      - command: attach.xunit_results
+        params:
+        files:
+          - "./stitch-java-sdk/android/services/twilio/build/outputs/androidTest-results/test/*.xml"
+
+task_groups:
```
adamchel: [nice] this will definitely help us debug builds more easily

.evg.yml
```diff
@@ -140,8 +146,37 @@ functions:
...
+          SDK_HOME=`pwd`/.android
+          export JAVA_HOME="/opt/java/jdk8"
+
+          export ANDROID_HOME=$SDK_HOME
+          export ANDROID_SDK_ROOT=$SDK_HOME
+          export ANDROID_SDK_HOME=$SDK_HOME
+
+          # Create the AVD
+          echo "Creating AVD"
+          echo no | $SDK_HOME/tools/bin/avdmanager create avd -n android_avd -c 1000M -k "system-images;android-26;google_apis;x86_64" -f
```
adamchel: [q] why Android 26?

.evg.yml
jsflax: There's only one macos-1014-stitch box. There are many macos-1014 boxes now. It allows us to parallelize our tasks (which is why the build time is so short now)

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/SyncWriteModelContainer.java
jsflax: I was testing to see if maybe there was a weird overhead in adding a `Set` to a `Document`. I can change it back, but I don't think it matters either way.

.evg.yml
adamchel: right, I guess what I was trying to say is that sometimes the server time will queue thousands of tasks, meaning that all of the macos-1014 boxes will have a potentially much longer queue than macos-1014-stitch. But I guess in general this is better, so we can definitely keep it

.evg.yml
jsflax: It's just what we had 🤷‍♂️

dkaminsky:
lgtm

PR #131, opened 2019-04-09
STITCH-2620 Java SDK: add hashing behavior to data synchronizer

dkaminsky:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -969,435 +895,571 @@ private void syncLocalToRemote() {
...
+                    committedEvent.getDocumentKey(),
+                    committedEvent.getUpdateDescription(),
+                    false));
+
+                setPendingWritesComplete = true;
+                if (committedEvent.getOperationType() != OperationType.DELETE) {
+                  localSyncWriteModelContainer.addConfigWrite(
+                      new ReplaceOneModel<>(CoreDocumentSynchronizationConfig.getDocFilter(
+                          nsConfig.getNamespace(), docConfig.getDocumentId()),
+                          docConfig));
```
adamchel: [q] it's a little unclear that the `docConfig` gets mutated later before the write is committed. Why don't we just call `docConfig.setPendingWritesComplete(nextVersion)` before the `addConfigWrite` call so it's clear that's what we're trying to convey with the local config write?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1477,13 +1557,8 @@ private SyncWriteModelContainer resolveConflict(
...
           transformedRemoteEvent);
     } catch (final Exception ex) {
-      logger.error(String.format(
-          Locale.US,
-          "t='%d': resolveConflict ns=%s documentId=%s resolution exception: %s",
-          logicalT,
-          namespace,
-          docConfig.getDocumentId(),
-          ex));
+      ex.printStackTrace(System.err);
```
adamchel: [nit] debugging artifact?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
+          final DocumentVersionInfo lastSeenVersionInfo =
+              DocumentVersionInfo.fromVersionDoc(docConfig.getLastKnownRemoteVersion());
+
+          final boolean remoteHasNoVersion = !remoteVersionInfo.hasVersion();
+          final boolean lastSeenHasNoVersion = !lastSeenVersionInfo.hasVersion();
+
+          final DocumentVersionInfo.Version remoteVersion =
+              remoteHasNoVersion ? null : remoteVersionInfo.getVersion();
+          final DocumentVersionInfo.Version lastSeenVersion =
+              lastSeenHasNoVersion ? null : lastSeenVersionInfo.getVersion();
```
jsflax: [nit] Both of these variables should be @Nullable

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
+          final BsonDocument remoteFullDocument = remoteChangeEvent.getFullDocument();
+
+          final long lastSeenHash;
+          if (lastSeenHasNoVersion && docConfig.getLastUncommittedChangeEvent() != null) {
+            // do a hash calculation if local is unversioned and we have a pending write
+            lastSeenHash =
+                HashUtils.hash(
+                    sanitizeDocument(
+                        docConfig.getLastUncommittedChangeEvent().getFullDocument()
+                    )
```
jsflax: [style] These three lines can fit on one line.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2927,4 +3018,96 @@ private static BsonDocument withNewVersion(
...
+    DELETE_AND_DESYNC(false, "; deleting and desyncing the document"),
+    DELETE_FROM_REMOTE(false, "; applying the remote delete"),
+    DROP(false, "; dropping the event"),
+    DROP_AND_PAUSE(false, "; dropping the event and pausing the document"),
+    DROP_AND_DESYNC(false, "; dropping the event and desyncing the document"),
+    WAIT(false, "; waiting until next pass"),
+    REMOTE_FIND(false, "; re-checking against remote collection"),
+    SYNTHETIC_CONFLICT(true, "; raising conflict");
+
+    private final boolean conflict;
```
jsflax: [nit] This field is never accessed. We should either remove it or explain why we should keep it.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2927,4 +3018,96 @@ private static BsonDocument withNewVersion(
...
+      return messageWithDescription;
+    }
+
+    @Override
+    public String toString() {
+      return this.message;
+    }
+  }
+
+  private enum SyncAction {
```
jsflax: [nit] We should consider moving this to its own file.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2927,4 +3018,96 @@ private static BsonDocument withNewVersion(
     newDocument.put(DOCUMENT_VERSION_FIELD, newVersion);
     return newDocument;
   }
+
```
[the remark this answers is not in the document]
jsflax: +1 – @tkaye407, please comment beneath the line you are commenting on, not above.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -969,435 +895,571 @@ private void syncLocalToRemote() {
...
 
     // 3. If there are still local writes pending for the document, it will go through the L2R
     //    phase on a subsequent pass and try to commit changes again.
+
   }
 
-  private void emitError(final CoreDocumentSynchronizationConfig docConfig,
-                         final String msg) {
-    this.emitError(docConfig, msg, null);
+  private LocalSyncWriteModelContainer performAction(
```
jsflax: [!important] There's a strong argument to be made that all of these parameters should be a part of a struct– maybe `SyncActionMessage`. A container would also make the flow within each method easier to follow.

[nit] This method doesn't necessarily perform an action– that verbiage can be confusing since it returns a `LocalSyncWriteModelContainer`. It should either have a more semantically appropriate name (`applyAction`, `processAction`) or have a docString explaining what's going on.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
    */
   @CheckReturnValue
   private @Nullable
-  SyncWriteModelContainer syncRemoteChangeEventToLocal(
+  LocalSyncWriteModelContainer syncRemoteChangeEventToLocal(
```
jsflax: [minor] The cyclomatic complexity of this method is fairly high at this point. Since breaking this method/class up is out of scope for now, I would really appreciate more line-by-line comments (akin to what Adam had previously).

jsflax:
Also can we get evergreen passing on this?

android/examples/stress-tests/build.gradle
```diff
@@ -36,7 +36,11 @@ dependencies {
...
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.1.1"
-
-
+    implementation ("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.1.1") {
+        exclude group: "org.jetbrains.kotlin", module: "kotlin-stdlib"
+        exclude group: "org.jetbrains.kotlin", module: "kotlin-stdlib-jdk8"
+    }
+    // instead of transitively getting wrong version from coroutines library
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
```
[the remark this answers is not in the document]
dkaminsky: yes, I did mean to add this

README.md
```diff
@@ -71,7 +71,6 @@ This will start you off with the core SDK functionality as well as the remote Mo
 For customized dependencies use the following:
 
 ```gradle
-implementation 'org.mongodb:stitch-server-core:4.2.1'
```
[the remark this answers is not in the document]
dkaminsky: I think I absorbed this change from an unrelated merge - will revert

android/services/mongodb-remote-perftests/src/main/AndroidManifest.xml
```diff
@@ -1 +0,0 @@
-<manifest package="com.mongodb.stitch.android.services.mongodb.performance" />
```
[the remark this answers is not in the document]
dkaminsky: +1 I restored it locally already during my own look at the PR, just haven't pushed it yet

android/examples/stress-tests/build.gradle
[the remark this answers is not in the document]
dkaminsky: It resolves dependency issues with the kotlin upgrade

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
+                message = SyncMessage.APPLY_FROM_REMOTE_MESSAGE;
+              } else {
+                if (lastSeenVersion.getSyncProtocolVersion() == SYNC_PROTOCOL_VERSION
+                    && lastSeenVersion.getInstanceId().equals(remoteVersion.getInstanceId())) {
+                  message = SyncMessage.PROBABLY_GENERATED_BY_US_MESSAGE;
+                  action = SyncAction.DROP;
+                } else if (remoteVersionCounter == lastSeenVersionCounter
+                    && lastSeenHash != remoteHash) {
+                  message = SyncMessage.REMOTE_UPDATE_WITHOUT_VERSION_MESSAGE;
+                  action = SyncAction.SYNTHETIC_CONFLICT;
```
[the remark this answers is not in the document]
dkaminsky: this handles the case of an external update from outside of sync (eg from web or a server application) where the value was changed but not the sync version. the conflict forces the user-specified logic to decide what to do with this (e.g. remote wins by default) -- hmm I guess it makes more sense for this to apply if different rather than conflict...

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
+                } else {
```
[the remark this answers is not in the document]
dkaminsky: I think it was working that way previously, let's discuss offline

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -969,435 +895,571 @@ private void syncLocalToRemote() {
...
+                    nextVersion =
+                        DocumentVersionInfo.getFreshVersionDocument(
+                            localChangeEvent.getFullDocument());
+
+                    // It's possible that we may insert after a delete happened and we didn't get a
+                    // notification for it. There's nothing we can do about this.
+
+                    // a. Insert document into remote database
+                    try {
+                      remoteColl.insertOne(
```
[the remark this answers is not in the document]
dkaminsky: Are you suggesting I store it on the doc config itself rather than as part of the version? That will suppress it being sent to the remote but it breaks the logical connection between the three attributes (guid, version, hash)

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: because we have a read lock at this point and setPendingWritesComplete requires a write lock

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: frankly I was never on-board with the Effective Java "use no more than 2, sometimes 3" parameters nonsense - creating extra unneeded objects to pass around adds more gc overhead which can impact your performance in unpredictable ways. I'm not married to it being either way I just don't tend to like to create random data classes that get used in exactly one spot

nit fixed

dkaminsky: changed to enqueueAction

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: yep - thanks

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
dkaminsky: going to call that out of scope for now, breaking this up is a whole other monster and enough refactoring happened here already

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: I'd be more inclined to say that we need to break this whole class up into about 6 different files, but I think that can wait for now considering I'm doing this under time crunch for a particular use case

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/LocalSyncWriteModelContainer.java
```diff
@@ -103,10 +118,19 @@ void wrapForRecovery(final Runnable callable) {
   void commitAndClear() {
     wrapForRecovery(() -> {
       localWrites.commitAndClear();
-      configs.commitAndClear();
+      configWrites.commitAndClear();
       remoteWrites.commitAndClear();
     });
 
+    if (true) { // temporary: change this to a success indicator
```
adamchel: [nit] if you're planning on updating this in a future ticket of work, we should add the ticket number here. we've been trying to avoid ticket-less TODOs in the code where possible.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
-      logger.debug(String.format(
-          Locale.US,
-          "t='%d': syncRemoteChangeEventToLocal ns=%s documentId=%s remote change event was "
-              + "generated by us; dropping the event",
-          logicalT,
-          nsConfig.getNamespace(),
-          docConfig.getDocumentId()));
-      return null;
-    }
+      if (docConfig.hasUncommittedWrites() && docConfig.getLastResolution() == logicalT) {
```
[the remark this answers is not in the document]
adamchel: Yeah, when we talk about sync cadence I can explain a bit more. Running the sync pass twice isn't something that users will know to do.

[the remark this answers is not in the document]
adamchel: There's nothing we need to do right now

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreRemoteMongoCollectionWriteModelContainer.java
```diff
@@ -44,21 +46,33 @@ public CoreRemoteMongoCollectionWriteModelContainer(
    * Commits the writes to the remote collection.
    */
   @Override
-  public void commit() {
+  public boolean commit() {
     final CoreRemoteMongoCollection<DocumentT> collection = getCollection();
     final List<WriteModel<DocumentT>> writeModels = getBulkWriteModels();
 
+    // define success as any one operation succeeding for now
```
adamchel: [q] why not make it based on all three operations succeeding?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
+        action = SyncAction.WAIT;
+        message = SyncMessage.SIMULTANEOUS_WRITES_MESSAGE;
+      }
 
+      if (logger.isDebugEnabled()) {
+        logger.debug(String.format(
+            Locale.US,
+            SyncMessage.BASE_MESSAGE.concat(" processing remote operation='%s'"),
```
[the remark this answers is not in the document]
adamchel: I still think it's a bit hard to read, because it's unclear what the other arguments in `String.format` are referring to unless you actually look at the `BASE_MESSAGE` string. In any case, this is pretty nitty and not that big of a deal

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
adamchel: Alright, in that case can you add a quick comment then explaining that the doc config will have its pending writes set as complete by the time the write model commits its writes? I could see a future change accidentally messing up this logic.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -682,266 +706,166 @@ private void syncRemoteToLocal() throws InterruptedException {
...
-                      nsConfig.getNamespace(),
-                      docConfig.getDocumentId(),
-                      docConfig.hasUncommittedWrites()));
-    }
 
+      if (action == null) {
+        if (remoteVersionInfo.hasVersion()
+            && remoteVersionInfo.getVersion().getSyncProtocolVersion() != SYNC_PROTOCOL_VERSION) {
+          action = SyncAction.DROP_AND_DESYNC;
+          message = SyncMessage.UNKNOWN_REMOTE_PROTOCOL_VERSION_MESSAGE;
```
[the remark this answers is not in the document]
adamchel: Yeah this was always the behavior, I think it's fine as it stands.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/LocalSyncWriteModelContainer.java
dkaminsky: no this has to change now, this is what is breaking those three tests that are still failing

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreRemoteMongoCollectionWriteModelContainer.java
dkaminsky: it's actually something we'll have to fix long term with the "container" logic Jason added, which is that previously when a single operation failed, we knew immediately and wouldn't get to the point where we emitted the event. Now, we can't tell individually what passes and what fails. This is problematic. I'd rather over-emit duplicated events than fail to emit a unique event.

jsflax: That seems arbitrary– there are plenty of workarounds for this, including returning multiple values here.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -969,435 +915,537 @@ private void syncLocalToRemote() {
...
+                // 2. Otherwise, the unprocessed event can be safely dropped and ignored in future
+                // R2L passes. Continue on to checking the operation type.
+              }
 
+              if (action == null) { // if we haven't encountered an error/conflict already
+                // iii. Check the operation type
+                switch (localChangeEvent.getOperationType()) {
+                  // 1. INSERT
+                  case INSERT: {
+                    nextVersion =
```
adamchel: [minor] looks like you're still computing and sending the hash as part of the remote version

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -969,435 +915,537 @@ private void syncLocalToRemote() {
...
+      DocumentVersionInfo newestRemoteVersionInfo = null;
+      try {
+        newestRemoteVersionInfo = DocumentVersionInfo.getRemoteVersionInfo(newestRemoteDocument);
+      } catch (final Exception e) {
+        action = SyncAction.DROP_EVENT_AND_DESYNC;
+        message = SyncMessage.CANNOT_PARSE_REMOTE_VERSION_MESSAGE;
+      }
+
+      if (newestRemoteVersionInfo != null) { // if no error has occurred
+        final boolean newestRemoteHasNoVersion = !newestRemoteVersionInfo.hasVersion();
```
adamchel: [nit] in this particular block of code, I think `newestRemoteHasVersion` and `lastSeenHasVersion` make more sense. 

This later on is a mouthful:
```
if (!newestRemoteHasNoVersion && !lastSeenHasNoVersion
```

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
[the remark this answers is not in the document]
dkaminsky: done

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: removed

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/LocalSyncWriteModelContainer.java
dkaminsky: changed

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreRemoteMongoCollectionWriteModelContainer.java
dkaminsky: it's a limitation in the result provided by Mongo

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: fixed

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
dkaminsky: agree to disagree for the sake of getting this done, feel free to refactor on next pass

jsflax:
Lgtm

PR #143, opened 2019-04-17
STITCH-2715 add support for watching streams with compact change events

adamchel:
This PR took a lot longer than expected due to the fact that the generics were tricky to get working correctly without causing breaking changes for the existing watch API. The end result for users is relatively clean and mostly follows what we had before, but could probably still be improved. I tried to take advantage of type erasure in more places so that user-facing types aren't so verbose.

In any case, I think this all will be worth revisiting when we rework Java’s “watch” into a more user-friendly listener approach. 

The return types for `watch` and `watchCompact` in each of the user-facing SDKs are as follows
``` java
// watch for android (type parameters changed slightly, should now be simpler)
 Task<ChangeStream<Task<ChangeEvent<DocumentT>>>>

// watch for Java server (type parameters changed slightly, should not be simpler)
Task<ChangeStream<ChangeEvent<DocumentT>>>

// watchCompact for android 
Task<ChangeStream<Task<CompactChangeEvent<DocumentT>>>>

// watchCompact for Java server
 Task<ChangeStream<CompactChangeEvent<DocumentT>>> 
```

As a note, some of the tests in this PR are not actually written by me, they were just moved around as part of the restructuring of change events. For this reason, they may not necessarily follow the “test one thing per test” principle. I did try to follow the principle when writing new tests, except for the two watch integration tests (happy to elaborate offline about why I did this).

The only other unaddressed thing in this PR was a weird discrepancy I was seeing between the way Android hashes object IDs vs how the backend server was hashing them. I’m going to investigate this further. I left a TODO and commented code in the watch integration tests where this was an issue, and I plan on addressing it before merging this PR.


android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/internal/RemoteMongoCollectionImpl.java
```diff
@@ -682,26 +683,54 @@ public ResultT call() {
...
+          }
+        }
+    );
+  }
+
+  @Override
+  public Task<ChangeStream<Task<CompactChangeEvent<DocumentT>>>> watchCompact(
+      final ObjectId... ids
+  ) {
+    return dispatcher.dispatchTask(
```
jsflax: [nit] We can do the conversation to BsonObjectIds here and pass them through to the overloaded method. This way, we can collapse the core implementation, and it doesn't have to be aware of the public API, as there is no need for it too.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/BaseChangeEvent.java
```diff
@@ -0,0 +1,110 @@
...
+
+import org.bson.BsonDocument;
+
+/**
+ * Represents the set of properties that exist on all MongoDB change events produced
+ * by watch streams in this SDK. Other change event types inherit from this type.
+ *
+ * @param <DocumentT> The type of the full document in the change event.
+ */
+public abstract class BaseChangeEvent<DocumentT> {
```
jsflax: [important nit] I know it's a big ask, but is there any way to accomplish what you've done here without exposing `BaseChangeEvent` publicly? To be honest, BaseChangeEvent only exists to de-dupe code– it doesn't actually serve a practical purpose and exposes unnecessary implementation details to the user.

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoCollection.java
```diff
@@ -474,17 +475,43 @@
...
+   * use case of non-{@link BsonValue} instances of {@link ObjectId}. Requests a stream where the
+   * full document of update events, and several other unnecessary fields are omitted from the
+   * change event objects returned by the server. This can save on network usage when watching
+   * large documents.
+   *
+   * @param ids unique object identifiers of the IDs to watch.
+   * @return the stream of change events.
+   */
+  Task<ChangeStream<Task<CompactChangeEvent<DocumentT>>>> watchCompact(final ObjectId... ids);
+
```
tkaye407: I know I'm late to this conversation but it seems like we are sacrificing usability for not breaking the current API. Theoretically, why can't we make ChangeEvent<DocumentT> an interface and return it in both of these methods? 

Also is it better to have another name for the method "watchCompact" as opposed to an overload of the watch method that is: watch(final bool fullDocument = false, final ObjectId... ids)

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/CompactChangeEvent.java
```diff
@@ -0,0 +1,204 @@
...
+
+    return new CompactChangeEvent<>(
+        OperationType.fromRemote(document.getString(Fields.OPERATION_TYPE_FIELD).getValue()),
+        fullDocument,
+        document.getDocument(Fields.DOCUMENT_KEY_FIELD),
+        updateDescription,
+        stitchDocumentVersion,
+        stitchDocumentHash,
+        document.getBoolean(Fields.WRITE_PENDING_FIELD, BsonBoolean.FALSE).getValue());
+  }
```
tkaye407: It would be interesting to add some kind of utility function here for people who want to use the compactChangeEvent and combine it with an existing document since we have that logic somewhere (ie applying the update / deleted fields).

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/internal/RemoteMongoCollectionImpl.java
adamchel: I can do that, but I think the reason it was in core was so that conversion logic only needed to happen in core, and not in both android and server.

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoCollection.java
adamchel: We can't make `ChangeEvent` an interface because the two change event models are fundamentally different. 

For that same reason, we have multiple methods. You can't overload on return type in Java, so if you have a different return type depending on the argument, you either have to have a generic return type which is not very user friendly, or we'd have to have this be a stream of `BaseChangeEvent`, which is not very useful to users since they would have to know to typecast to the type of change event they wany. Happy to clarify offline as well.

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/internal/RemoteMongoCollectionImpl.java
dkaminsky: This was the reason +1

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/BaseChangeEvent.java
dkaminsky: I’m not terribly particular about publicly exposing internal classes in Java. It’s a well-known shortcoming of their inheritance scheme that often things need to be made public so that they remain visible to other parts of the code even though they are not intended to be consumed by users

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/CompactChangeEvent.java
dkaminsky: I’m not sure what the utility of that would be at the moment, unless we had another source of change events coming in...

tkaye407: We chatted about this offline, but say you have a Todo item and one of its fields is updated and you want to limit your network usage so you use a consiseChangeEvent. I was just suggesting that the consciseChangeEvent could have a method that takes an existing document and applied the changes in updateFields and deletedFields to the given document.

PR #148, opened 2019-04-23
STITCH-2843 Java SDK: Fix Update => Update, Update => Replace, Replace => Update Coalescence

jsflax:


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/CoreDocumentSynchronizationConfig.java
```diff
@@ -417,6 +417,48 @@ public long getLastKnownHash() {
...
+                newestChangeEvent.hasUncommittedWrites()
+            );
+        }
+      case REPLACE:
+        switch (newestChangeEvent.getOperationType()) {
+          case UPDATE:
+            return new ChangeEvent<>(
+                newestChangeEvent.getId(),
+                OperationType.REPLACE,
+                newestChangeEvent.getFullDocument(),
```
dkaminsky: [q] does a local update always have the full document in the change event?

jsflax: Yes, I cannot think of a case where we do not have the full document.

PR #152, opened 2019-04-29
STITCH-2718 use update description deltas in mobile sync

adamchel:
This is the big one. This updates the mobile sync codebase to use `watch` stream with compact events rather than traditional MongoDB change events. This required us to modify all code that operated on the assumption that `UPDATE` change events would have full documents.

This means that the DataSynchronizer now operates with `CompactChangeEvent`s when working with remote change events, and `ChangeEvent`s when working with local change events.

This meant changing all the unit tests and integration tests to use the new conflict handler, and to have remote events that are compact change events rather than full change events.

All unit tests and integration tests are passing locally, so any evergreen failure is likely to be an OOM error.

Notable Drive-Bys:
* The user-defined custom conflict handler now needs to return a `ConflictResolution` rather than a full document. This paves the way for defining richer conflict resolution cases in the future, and lets the data synchronizer figure out how to handle common cases like from remote or from local which can no longer be done with just the remote event.
    * This will be a major breaking change to the sync API
* REPLACE -> UPDATE coalescence now works by applying the UPDATE's update description to the REPLACE's full document rather than taking the UPDATE's full document.

### DIFF
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ConflictResolution.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ConflictResolution.java
@@ -1,0 +1,69 @@
+package com.mongodb.stitch.core.services.mongodb.remote.sync;
+
+public abstract class ConflictResolution {
+
+  public enum ConflictResolutionType {
+    WITH_DOCUMENT,
+    FROM_LOCAL,
+    FROM_REMOTE
+  }
+
+  public static final class WithDocument<T> extends ConflictResolution {
+    private final T fullDocumentForResolution;
+
+    WithDocument(final T fullDocumentForResolution) {
+      this.fullDocumentForResolution = fullDocumentForResolution;
+    }
+
+    @Override
+    public ConflictResolutionType getType() {
+      return ConflictResolutionType.WITH_DOCUMENT;
+    }
+
+    public T getFullDocumentForResolution() {
+      return fullDocumentForResolution;
+    }
+  }
+
+  public static final class FromRemote extends ConflictResolution {
+
+    static final FromRemote instance = new FromRemote();
+
+    private FromRemote() {
+    }
+
+    @Override
+    public ConflictResolutionType getType() {
+      return ConflictResolutionType.FROM_REMOTE;
+    }
+
+  }
+
+  public static final class FromLocal extends ConflictResolution {
+
+    static final FromLocal instance = new FromLocal();
+
+    private FromLocal() {
+    }
+
+    @Override
+    public ConflictResolutionType getType() {
+      return ConflictResolutionType.FROM_LOCAL;
+    }
+
+  }
+
+  public static <T> ConflictResolution withDocument(T fullDocumentForResolution) {
+    return new WithDocument<>(fullDocumentForResolution);
+  }
+
+  public static ConflictResolution fromRemote() {
+    return FromRemote.instance;
+  }
+
+  public static ConflictResolution fromLocal() {
+    return FromLocal.instance;
+  }
+
+  public abstract ConflictResolutionType getType();
+}
--- a/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ConflictResolution.java
+++ b/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ConflictResolution.java
@@ -1,13 +1,44 @@
+/*
+ * Copyright 2018-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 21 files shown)


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -2562,6 +2692,81 @@ private LocalSyncWriteModelContainer replaceOrUpsertOneFromRemote(
...
+      final NamespaceSynchronizationConfig nsConfig,
+      final BsonValue documentId,
+      final UpdateDescription updateDescription,
+      final BsonDocument atVersion,
+      final long withHash
+  ) {
+    if (updateDescription.isEmpty()) {
+      // don't do anything for completely no-op updates
+      return null;
+    }
```
tkaye407: Was thinking you may want to do that after the call to sanitizeUpdateDescription()

core/services/mongodb-remote/src/test/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizerCRUDUnitTests.kt
```diff
@@ -47,13 +46,23 @@ class DataSynchronizerCRUDUnitTests {
...
             // do a sync pass, accepting the local doc. this will create
             // a pending replace to be sync'd on the next pass
-            ctx.queueConsumableRemoteUpdateEvent()
+            ctx.queueConsumableRemoteUpdateEvent(
+                ctx.testDocumentId,
+                ctx.testDocumentVersion,
+                ctx.testDocumentHash,
+                UpdateDescription(BsonDocument(), HashSet()),
+                TestVersionState.NEXT
+
```
tkaye407: nit: Get rid of this line

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
```diff
@@ -1856,6 +1864,12 @@ class SyncIntTestProxy(private val syncTestRunner: SyncIntTestRunner) {
 
     private fun waitForAllStreamsOpen() {
         while (!syncTestRunner.dataSynchronizer.areAllStreamsOpen()) {
+            if (syncTestRunner
+                    .dataSynchronizer
+                    .getSynchronizedDocuments(syncTestRunner.namespace).size == 0) {
+                println("there are no documents to watch, so the stream will never open")
+                break
+            }
```
tkaye407: Good catch here, surprised we never had tests that got tripped up on this

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
```diff
@@ -1352,23 +1375,79 @@ private LocalSyncWriteModelContainer enqueueAction(
...
+              docConfig.getDocumentId(),
+              remoteChangeEvent.getUpdateDescription(),
+              applyNewVersion,
+              remoteHash
+          );
+        }
+
+        if (writeContainer == null) {
+          return null;
+        }
```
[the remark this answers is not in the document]
adamchel: As discussed offline, this is necessary because `updateOneFromRemote` can return null in the case of no-op updates.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/DataSynchronizer.java
adamchel: As discussed offline, this needs to be here since an empty sanitized update description might still have version changes that make it not a no-op.

core/testutils/src/main/java/com/mongodb/stitch/core/testutils/sync/SyncIntTestProxy.kt
adamchel: yeah, we actually had tests that would hang once in a while, but this should resolve that

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ConflictResolution.java
```diff
@@ -0,0 +1,134 @@
...
+ * {@link ConflictResolution} represents a resolution to a MongoDB Mobile Sync conflict. Conflicts
+ * occur when a local pending write conflicts with an incoming remote change. The user-defined
+ * conflict handler method in a {@link ConflictHandler} returns a {@link ConflictResolution}.
+ */
+public abstract class ConflictResolution {
+
+  /**
+   * The different ways of resolving a conflict.
+   */
+  public enum ConflictResolutionType {
```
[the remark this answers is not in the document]
adamchel: I wasn't able to make all of these package private since they're necessary in the DataSynchronizer, which is in the internal package. As Doug said, this is just the nature of Java so we won't be able to completely abstract it away. I added "For internal use only" in the doc comments where relevant.

adamchel:
@jsflax PTAL

PR #153, opened 2019-04-30
STITCH-2847 - delays event emission until local changes are applied

dkaminsky:
Addresses the case where sometimes events are received before the local changes are applied by deferring the emission of the event to the write model container in a few places in the data synchronizer.


tkaye407:
Dont worry about the checkstyle issue I think I fix it in my current PR

PR #158, opened 2019-05-10
STITCH-2785 Java SDK: Permit user to specify a callback or callback/executor for watch

tkaye407:
Adds the listener pattern for watch() where watch() returns a ChangeStream that can then take multiple ChangeStreamListeners and run onEvent() for each of them on the next event.


core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
```diff
@@ -0,0 +1,80 @@
...
+        boolean hasListeners = false;
+        while (listeners.hasMoreElements()) {
+          listeners.nextElement().onEvent(nextEvent.getDocumentKey(), nextEvent);
+          hasListeners = true;
+        }
+
+        if (!hasListeners) {
+          Thread.currentThread().interrupt();
+        }
+      } catch (Exception e) {
```
tkaye407: Not sure what the best behaviour is here

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1085,294 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+        rawDoc2["_id"] = objectId2
+        rawDoc2["happy"] = "day"
+
+        Tasks.await(coll.insertMany(listOf(rawDoc1, rawDoc2)))
+        assertEquals(2, Tasks.await(coll.count()))
+
+        val streamTask = coll.watch(objectId1, objectId2)
+        val stream = Tasks.await(streamTask)
+
+        var changeEventListenerSemaphore = Semaphore(0)
```
dkaminsky: This Semaphore + AtomicInteger mechanic can likely be replaced with a CountdownLatch

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1085,294 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+        rawDoc2["_id"] = objectId2
+        rawDoc2["happy"] = "day"
+
+        Tasks.await(coll.insertMany(listOf(rawDoc1, rawDoc2)))
+        assertEquals(2, Tasks.await(coll.count()))
+
+        val streamTask = coll.watch(objectId1, objectId2)
+        val stream = Tasks.await(streamTask)
+
+        var changeEventListenerSemaphore1 = Semaphore(0)
```
dkaminsky: ditto

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1085,294 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+        rawDoc2["_id"] = objectId2
+        rawDoc2["happy"] = "day"
+
+        Tasks.await(coll.insertMany(listOf(rawDoc1, rawDoc2)))
+        assertEquals(2, Tasks.await(coll.count()))
+
+        val streamTask = coll.watchCompact(objectId1, objectId2)
+        val stream = Tasks.await(streamTask)
+
+        var changeEventListenerSemaphore1 = Semaphore(0)
```
dkaminsky: ditto

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1085,294 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+        rawDoc1["_id"] = objectId1
+        rawDoc1["hello"] = "world"
+
+        Tasks.await(coll.insertOne(rawDoc1))
+        assertEquals(1, Tasks.await(coll.count()))
+
+        val streamTask = coll.watch(objectId1)
+        val stream = Tasks.await(streamTask)
+
+        var changeEventListenerSemaphore1 = Semaphore(0)
```
dkaminsky: ditto

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/AsyncChangeStream.java
```diff
@@ -35,49 +35,94 @@
...
+      final BaseChangeEventListener<DocumentT, ChangeEventT> listener) {
+    stream.addChangeEventListener(listener);
+  }
+
+  /**
+   * Remove a ChangeEventListener from the ChangeStream.
+   *
+   * @param listener the ChangeEventListener
+   */
+  public void removeChangeEventListener(
```
dkaminsky: I still think exposing BaseChangeEventListener to the user doesn't feel like the right level of abstraction.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
```diff
@@ -70,23 +117,51 @@ public boolean isOpen() {
...
     internalStream.close();
+    runnerThread = null;
   }
 
-  protected Stream<? extends BaseChangeEvent> getInternalStream() {
+  protected void removeAllChangeEventListeners() {
+    listeners.clear();
+  }
+
+  protected Enumeration<BaseChangeEventListener> getChangeEventListeners() {
```
dkaminsky: ditto - really don't like BaseCEL as the user-level abstraction

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
dkaminsky: Probably catch InterruptedException explicitly, that's a no-op, any other exception maybe just throw an exception and log an error?

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
```diff
@@ -0,0 +1,80 @@
...
+    if (stream == null) {
+      return;
+    }
+
+    // not sure actually if we should do this
+    stream.removeAllChangeEventListeners();
+
+    stream.close();
+  }
+}
```
dkaminsky: [nit] missing newline at end of file

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/AsyncChangeStream.java
tkaye407: How so? I tested to make sure that this is type safe in that you cannot send in the wrong type of listener.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
tkaye407: This isnt exposed to users

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
```diff
@@ -0,0 +1,79 @@
...
+        while (listeners.hasMoreElements()) {
+          listeners.nextElement().onEvent(nextEvent.getDocumentKey(), nextEvent);
+          hasListeners = true;
+        }
+
+        if (!hasListeners) {
+          Thread.currentThread().interrupt();
+        }
+      } catch (Exception e) {
+        Loggers.getLogger("ChangeStreamListener").error(
```
adamchel: [minor] you should definitely dispatch this exception to the listeners

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
```diff
@@ -0,0 +1,79 @@
...
+            "Change Stream Listener will terminate due to exception: " + e.getLocalizedMessage());
+        Thread.currentThread().interrupt();
+      }
+    } while (!Thread.currentThread().isInterrupted());
```
adamchel: [major?] does this thread actually stop when you call `close` on the stream? seems like it will be hang on nextEvent because nothing except an error will interrupt this thread.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
```diff
@@ -70,23 +117,43 @@ public boolean isOpen() {
   }
 
   /**
-   * Closes the underlying stream.
+   * Indicates whether or not any ChangeStreamListeners are currently running.
+   * @return True if the ChangeStreamListeners are running
+   */
+  public boolean listenersRunning() {
```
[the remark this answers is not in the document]
adamchel: though I see that if the thread is already running, you don't want to recreate it. I think you should have this method be named `listenersThreadRunning` and make it internal and not intended to be used by users. A listener doesn't run, the thread that dispatches events to the listeners is running.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
adamchel: if the problem with dispatching this exception is that closing the stream causes an IOException, we need a cleaner way to handle the case where the stream is closing.

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/ChangeEventListener.java
```diff
@@ -17,14 +17,14 @@
...
 import com.mongodb.stitch.core.services.mongodb.remote.ChangeEvent;
-
 import org.bson.BsonValue;
 
 /**
  * ChangeEventListener receives change event notifications.
  * @param <DocumentT> the type of class represented by the document in the change event.
  */
-public interface ChangeEventListener<DocumentT> {
+public interface ChangeEventListener<DocumentT>
```
adamchel: [nit] this is clean, but not clear to users since they shouldn't need to have to look into an internal type to see what the interface should look like. Do what you did with `CompactChangeEventListener` and just add the method signature.

server/services/mongodb-remote/src/test/java/com/mongodb/stitch/server/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -18,6 +18,8 @@ import com.mongodb.stitch.server.testutils.BaseStitchServerIntTest
 import com.mongodb.stitch.core.services.mongodb.remote.RemoteFindOptions
```
adamchel: same comments in this file as the android tests

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1084,262 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+
+        val latch2 = CountDownLatch(2)
+        val listener2 = ChangeEventListener<Document> { _, event ->
+            assertNotNull(event.fullDocument)
+            latch2.countDown()
+        }
+
+        stream.addChangeEventListener(listener1)
+        stream.addChangeEventListener(listener2)
+        stream.addChangeEventListener(listener1)
```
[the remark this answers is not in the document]
tkaye407: just testing to make sure it's not adding another listener

android/services/mongodb-remote/src/androidTest/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -1080,6 +1084,262 @@ class RemoteMongoClientIntTests : BaseStitchAndroidIntTest() {
...
+            assertEquals(latch1.count.toInt(), 2)
+
+            stream.removeChangeEventListener(listener2)
+
+            Tasks.await(coll.updateMany(BsonDocument(),
+                Document().append("\$set", Document().append("new", "field2"))))
+        } finally {
+            stream.close()
+        }
+        assert(latch1.await(10, TimeUnit.SECONDS))
```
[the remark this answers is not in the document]
tkaye407: ill put it in the try

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/AsyncChangeStream.java
```diff
@@ -35,48 +35,93 @@
...
+    stream.close();
+  }
+
+  /**
+   * Optionally adds a listener that is notified when an attempt to retrieve the next event.
+   * fails.
+   *
+   * @param exceptionListener The exception listener.
+   */
+  public void setExceptionListener(final ExceptionListener exceptionListener) {
```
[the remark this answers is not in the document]
tkaye407: I dont actually trigger any errors, nextEvent does which it already was doing before so nothing is changing here. Will add if though if youd like

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/RemoteMongoCollection.java
```diff
@@ -479,15 +478,15 @@
    * @param ids unique object identifiers of the IDs to watch.
    * @return the stream of change events.
    */
-  Task<ChangeStream<Task<ChangeEvent<DocumentT>>>> watch(final ObjectId... ids);
+  Task<AsyncChangeStream<DocumentT, ChangeEvent<DocumentT>>> watch(final ObjectId... ids);
```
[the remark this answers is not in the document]
tkaye407: Yes I know that, but I couldnt figure out any other way to do it. However, I would doubt many people are using watch in its current form

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStreamRunner.java
tkaye407: it will get interrupted when the stream calls close now, good catch

android/services/mongodb-remote/src/main/java/com/mongodb/stitch/android/services/mongodb/remote/AsyncChangeStream.java
```diff
@@ -103,21 +105,18 @@ public boolean isOpen() {
...
-    return stream.listenersRunning();
+  public boolean areListenersAttached() {
+    return stream.areListenersAttached();
   }
 
   /**
    * Returns a {@link Task} whose resolution gives the next event from the underlying stream.
    * @return task providing the next event
    * @throws IOException if the underlying stream throws an {@link IOException}
+   * @throws IllegalStateException if any ChangeEventListeners are running
```
adamchel: seems like you're no longer throwing the exception?

adamchel: ah nvm it seems like you moved it

core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/ChangeStream.java
```diff
@@ -120,11 +135,11 @@ public boolean isOpen() {
    * Indicates whether or not any ChangeStreamListeners are currently running.
    * @return True if the ChangeStreamListeners are running
    */
-  public boolean listenersRunning() {
+  public boolean areListenersAttached() {
     if (runnerThread == null) {
       return false;
     }
-    return runnerThread.isAlive();
```
adamchel: [minor] this is super unlikely to happen, but this introduces the issue where if a listener is added (creating the runnerThread thread), but then removed (so the runner thread is still alive but listeners is empty), then nextEvent will not throw the IllegalStateException, even though it should because the runnerThread already called it.

I think this method should be called "isListenerThreadRunning", and it should just return `runnerThread.isAlive()`. It can be private or package private, since a user doesn't really need this method.

tkaye407:
Have all of the LGTM's and ran all of the tests as well as ktlint, checkstyle and findbugs locally. Evergreen is broken, so I will be merging it into master even though it is failing.

PR #170, opened 2019-07-04
STITCH-2988 Support additional watch options in Android SDK

adamchel:
This adds support for full-collection and filter-based watch. I was not able to test ktlint locally, will only merge if Evergreen passes.

### DIFF
--- a/android/core/src/main/AndroidManifest.xml
+++ b/android/core/src/main/AndroidManifest.xml
@@ -3,7 +3,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /><uses-permission android:name="android.permission.INTERNET" />
 
-    <application android:usesCleartextTraffic="true">
+    <application>
         <!-- StitchInitProvider is a ContentProvider that will associate the application's Context
           with Stitch. It also will look for resource values such as stitch_client_app_id to register
            a default app client.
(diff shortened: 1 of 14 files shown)


android/core/src/main/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /><uses-permission android:name="android.permission.INTERNET" />
 
-    <application android:usesCleartextTraffic="true">
```
tkaye407: :|

server/services/mongodb-remote/src/test/java/com/mongodb/stitch/server/services/mongodb/remote/RemoteMongoClientIntTests.kt
```diff
@@ -812,6 +812,145 @@ class RemoteMongoClientIntTests : BaseStitchServerIntTest() {
...
+            assertEquals(1, coll.count())
+
+            coll.insertOne(rawDoc2)
+            assertEquals(2, coll.count())
+
+            coll.updateMany(BsonDocument(), Document().append("\$set",
+                Document().append("new", "field")))
+
+            // should be filtered
+            // val insertEvent1 = stream.nextEvent()
```
tkaye407: ?

adamchel: basically just making it clear that if you tried to fetch the next event from the stream, it shouldn't be that one because the match expression should filter it

PR #179, opened 2019-10-17
STITCH-3383 add client sdk FunctionCredential

jsflax:


android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAppClientIntTests.kt
```diff
@@ -304,4 +308,39 @@ class StitchAppClientIntTests : BaseStitchAndroidIntTest() {
...
+            private = false
+        ))
+
+        addProvider(app.second, ProviderConfigs.Function(function.id, function.name))
+
+        val client = getAppClient(app.first)
+
+        val user = Tasks.await(
+            client.auth.loginWithCredential(
+                FunctionCredential(Document("id", "123abc"))))
```
mburdette: [nit] do this in a `beforeEach` do each assertion below in its own it block with good description for each

core/sdk/src/main/java/com/mongodb/stitch/core/auth/providers/function/FunctionCredential.java
```diff
@@ -0,0 +1,66 @@
...
+    return providerName;
+  }
+
+  @Override
+  public String getProviderType() {
+    return FunctionAuthProvider.TYPE;
+  }
+
+  @Override
+  public Document getMaterial() {
```
mburdette: why is this called `getMaterial` ?

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAppClientIntTests.kt
```diff
@@ -304,4 +308,39 @@ class StitchAppClientIntTests : BaseStitchAndroidIntTest() {
...
+
+    @Test
+    fun testCustomFunctionLogin() {
+        val app = createApp()
+
+        val function = app.second.functions.create(FunctionCreator(
+            name = "funkyAuth",
+            source = """
+            exports = function(payload) {
+                return "foo";
```
mburdette: [nit] might be good to have the function actually use the payload in some way?

core/sdk/src/main/java/com/mongodb/stitch/core/auth/providers/function/FunctionCredential.java
jsflax: Simply the name of the function we've always used. Each credential has associated metadata, but metadata is an overloaded term– so it provides material to the auth service.

android/coretest/src/androidTest/java/com/mongodb/stitch/android/core/StitchAppClientIntTests.kt
jsflax: Not sure why this would be in a beforeEach block, we're only doing it in one integration test.

